PR #3, opened 2019-10-03
fix: refresh access token if the JWT cookie has expired

douglashall:


src/learnerPortalLinks.js
```diff
@@ -60,7 +60,11 @@ function getCachedLearnerPortalLinks(userId) {
 
 export default async function getLearnerPortalLinks(apiClient) {
   let learnerPortalLinks = [];
-  const accessToken = apiClient.getDecodedAccessToken();
+  let accessToken = apiClient.getDecodedAccessToken();
+  if (!accessToken) {
+    await apiClient.refreshAccessToken();
+    accessToken = apiClient.getDecodedAccessToken();
+  }
```
nasthagiri: Curious: is there an issue in `frontend-auth` that's not automatically refreshing the token?  This seems like something others may run into and it would be better to fix `frontend-auth` if needed.

douglashall: No, `frontend-auth` is not broken. The use case here is to inspect the JWT's roles claim to determine whether or not a user is associated with an enterprise. The original code here had a bug in that it did not consider the case when the JWT cookie has expired. This fixes that mistake.

PR #8, opened 2020-02-25
ENT-2646 | Get learner portal hostname from env vars instead of api call

christopappas:


src/learnerPortalLinks.js
```diff
@@ -19,17 +19,21 @@ function cacheLinks(userId, links) {
...
       const enterpriseCustomer = enterpriseCustomers[i];
+      const enterpriseCustomerSlug = enterpriseCustomer.slug;
       const enableLearnerPortal = enterpriseCustomer.enable_learner_portal;
-      const learnerPortalHostname = enterpriseCustomer.learner_portal_hostname;
-      if (enableLearnerPortal && learnerPortalHostname) {
+      if (enableLearnerPortal && enterpriseCustomerSlug) {
         learnerPortalLinks.push({
           title: `${enterpriseCustomer.name}`,
-          url: `https://${enterpriseCustomer.learner_portal_hostname}`,
+          url: `https://${learnerPortalHostname}/${enterpriseCustomerSlug}`,
```
adamstankiewicz: I wonder if we should dynamically determine the HTTP scheme/protocol (http vs. https) for this URL, so that when working locally, it uses http but https on stage/prod.

We should be able to get the HTTP scheme/protocol from `window.location.protocol`.

christopappas: sounds good to me

src/learnerPortalLinks.js
```diff
@@ -19,17 +19,21 @@ function cacheLinks(userId, links) {
...
       const enterpriseCustomer = enterpriseCustomers[i];
+      const enterpriseCustomerSlug = enterpriseCustomer.slug;
       const enableLearnerPortal = enterpriseCustomer.enable_learner_portal;
-      const learnerPortalHostname = enterpriseCustomer.learner_portal_hostname;
-      if (enableLearnerPortal && learnerPortalHostname) {
+      if (enableLearnerPortal && enterpriseCustomerSlug) {
         learnerPortalLinks.push({
           title: `${enterpriseCustomer.name}`,
-          url: `https://${enterpriseCustomer.learner_portal_hostname}`,
+          url: `${window.location.protocal}//${learnerPortalHostname}/${enterpriseCustomerSlug}`,
```
adamstankiewicz: Typo: "protocal" --> "protocol"

christopappas: 😬 
thank you

PR #11, opened 2020-07-07
ENT-2893: Adds branding configuration to links

jlajoie:
Example outputs:
```
[
  {
    title: 'Test Enterprise',
    url: 'https://localhost:18000/test-enterprise',
    branding_configuration: {
      logo: 'http://localhost:18000/media/enterprise/branding/None/None_logo.png',
      banner_border_color: '#bbbbbb',
      banner_background_color: '#eeeeee'
    }
  }
]
```
```
[
  {
    title: 'Test Enterprise',
    url: 'https://localhost:18000/test-enterprise',
    branding_configuration: null
  }
]
```

### DIFF
--- a/src/learnerPortalLinks.js
+++ b/src/learnerPortalLinks.js
@@ -29,10 +29,18 @@ async function fetchLearnerPortalLinks(apiClient, userId) {
       const enterpriseCustomer = enterpriseCustomers[i];
       const enterpriseCustomerSlug = enterpriseCustomer.slug;
       const enableLearnerPortal = enterpriseCustomer.enable_learner_portal;
+      const brandingConfiguration = enterpriseCustomer.branding_configuration;
       if (enableLearnerPortal && enterpriseCustomerSlug) {
         learnerPortalLinks.push({
           title: `${enterpriseCustomer.name}`,
           url: `${window.location.protocol}//${learnerPortalHostname}/${enterpriseCustomerSlug}`,
+          // branding_configuration is not always returned as part
+          // of the response so check if it exists before referencing fields
+          branding_configuration: brandingConfiguration ? {
+            logo: brandingConfiguration.logo || null,
+            banner_border_color: brandingConfiguration.banner_border_color || null,
+            banner_background_color: brandingConfiguration.banner_background_color || null,
+          } : null,
         });
       }
     }


binodpant:
From the little I know of this particular repo, this looks good to me. Can vaguely associate this with the branding work I did in learner portal enterprise. 

Do any tests need to be updated/edited as part of this? Will this change be tested as part of another PR when it is subsumed? Perhaps it is sufficient to update testing notes even, don't know. just curious.

src/learnerPortalLinks.js
```diff
@@ -29,10 +29,18 @@ async function fetchLearnerPortalLinks(apiClient, userId) {
...
         learnerPortalLinks.push({
           title: `${enterpriseCustomer.name}`,
           url: `${window.location.protocol}//${learnerPortalHostname}/${enterpriseCustomerSlug}`,
+          // branding_configuration is not always returned as part
+          // of the response so check if it exists before referencing fields
+          branding_configuration: brandingConfiguration ? {
+            logo: brandingConfiguration.logo,
+            banner_border_color: brandingConfiguration.banner_border_color,
+            banner_background_color: brandingConfiguration.banner_background_color,
+          } : null,
```
binodpant, line 43: When I did work on learner portal enterprise I think I also had a case where brandingConfiguration was non null but some of the fields were, so I had to default them.. I do not know if that is a requirement here, is it?

for ref: https://github.com/edx/frontend-app-learner-portal-enterprise/blob/35ae057a7999a23f07d8a70c7a7acc1e9dfc3409/src/components/enterprise-page/data/hooks.js#L35

jlajoie: Great catch, realizing that all three of these fields are actually nullable so in order to keep the keys present I'll add a fallback.

jlajoie:
> From the little I know of this particular repo, this looks good to me. Can vaguely associate this with the branding work I did in learner portal enterprise.
> 
> Do any tests need to be updated/edited as part of this? Will this change be tested as part of another PR when it is subsumed? Perhaps it is sufficient to update testing notes even, don't know. just curious.

No tests in this repo, but I did verify the changes locally, I'll also be doing some smoke tests with this once landed to build out the header components so this'll be tested there.

PR #12, opened 2020-07-10
feat: adds function to return the preferred enterprise uuid for a user

jlajoie:
Used to get the UUID from a user's roles that match `enterprise_learner` as the first matching one is the selected and active enterprise for that user.

Usage example:
```
getLearnerPortalLinks(httpClient, authenticatedUser).then((learnerPortalLinks) => {
  const preferredUUID = getPreferredEnterpriseUUID(authenticatedUser);
  const preferredLearnerPortalLink = learnerPortalLinks.find(learnerPortalLink => learnerPortalLink.uuid == preferredUUID);
});
```

### DIFF
--- a/src/learnerPortalLinks.js
+++ b/src/learnerPortalLinks.js
@@ -1,5 +1,5 @@
 import { fetchEnterpriseCustomers } from './service';
-import { isEnterpriseLearner } from './utils';
+import { isEnterpriseUser } from './utils';
 
 function getCacheKey(userId) {
   return `learnerPortalLinks:${userId}`;
@@ -32,8 +32,6 @@ async function fetchLearnerPortalLinks(apiClient, userId) {
       const brandingConfiguration = enterpriseCustomer.branding_configuration;
       if (enableLearnerPortal && enterpriseCustomerSlug) {
         learnerPortalLinks.push({
-          title: `${enterpriseCustomer.name}`,
-          url: `${window.location.protocol}//${learnerPortalHostname}/${enterpriseCustomerSlug}`,
           // branding_configuration is not always returned as part
           // of the response so check if it exists before referencing fields
           branding_configuration: brandingConfiguration ? {
@@ -41,6 +39,9 @@ async function fetchLearnerPortalLinks(apiClient, userId) {
             banner_border_color: brandingConfiguration.banner_border_color || null,
             banner_background_color: brandingConfiguration.banner_background_color || null,
           } : null,
+          title: enterpriseCustomer.name,
+          url: `${window.location.protocol}//${learnerPortalHostname}/${enterpriseCustomerSlug}`,
+          uuid: enterpriseCustomer.uuid,
         });
       }
     }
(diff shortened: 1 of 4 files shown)


src/learnerPortalLinks.js
```diff
@@ -32,15 +32,16 @@ async function fetchLearnerPortalLinks(apiClient, userId) {
...
-          title: `${enterpriseCustomer.name}`,
-          url: `${window.location.protocol}//${learnerPortalHostname}/${enterpriseCustomerSlug}`,
           // branding_configuration is not always returned as part
           // of the response so check if it exists before referencing fields
           branding_configuration: brandingConfiguration ? {
             logo: brandingConfiguration.logo || null,
             banner_border_color: brandingConfiguration.banner_border_color || null,
             banner_background_color: brandingConfiguration.banner_background_color || null,
           } : null,
+          title: `${enterpriseCustomer.name}`,
```
adamstankiewicz, line 42: nit: idk why we had `enterpriseCustomer.name` in the backticks here haha. Can probably remove those.

src/utils.js
```diff
@@ -11,3 +11,19 @@ export function isEnterpriseLearner(user) {
 
   return false;
 }
+
+// eslint-disable-next-line import/prefer-default-export
```
adamstankiewicz, line 15: nit: with 2 named exports in this file now, we can remove the eslint disabling! 😄

src/utils.js
```diff
@@ -11,3 +11,19 @@ export function isEnterpriseLearner(user) {
...
+export function getPreferredEnterpriseUUID(user) {
+  if (user !== null && user.roles) {
```
adamstankiewicz, line 17: so i just learned about a JS operator through doing a PR review that helps with these sorts of conditionals. try `if (user?.roles)`! 😄

src/utils.js
```diff
@@ -11,3 +11,19 @@ export function isEnterpriseLearner(user) {
...
+    const { roles } = user;
+    for (let i = 0; i < roles.length; i += 1) {
```
adamstankiewicz, line 19: nit: might be able to do something like this in place of a `for` loop:

```
const filteredRoles = roles.filter(role => role.includes('enterprise_learner'));
const activeEnterpriseSelection = filteredRoles[0].split(':').pop();
return activeEnterpriseSelection;
```

src/utils.js
```diff
@@ -11,3 +11,19 @@ export function isEnterpriseLearner(user) {
...
+      if (roles[i].includes('enterprise_learner')) {
```
adamstankiewicz, line 20: I'm also wondering if these functions should be tied specifically to the `enterprise_learner` role like they are now, or if it should be more flexible to handle all role types: learners, admins, or openedx operators.

src/utils.js
jlajoie: Umm what?! That's hype.

src/utils.js
jlajoie: Oh excellent point, could pass that in as a param and make a set of constants available (if they're not already).

Might punt on this just in the short term because I'd rather things be consistent and I'm not looking to fix the isEnterpriseLearner method and everything downstream that calls it.

Unless you're saying we just change both of these methods to include every enterprise role type?

src/utils.js
jlajoie: And also unavailable until we swap to webpack 5 ☹️

src/utils.js
adamstankiewicz: Yeah, I'm suggesting making both function to be more generic to answer the question... "is this user Enterprise-related (any enterprise role)?" and, if so, "give that user their 'preferred' learner portal link".

Mainly suggesting because I've heard some feedback that admins might want to be using the Learner Portal as well (won't just be for learners). And if that's case, the navigation links from edx.org to LP should probably work for all role types.

PR #13, opened 2020-08-04
Rename new branding config field names

iloveagent57:
refer to the new primary/secondary/tertiary branding configuration colors.

### DIFF
--- a/src/learnerPortalLinks.js
+++ b/src/learnerPortalLinks.js
@@ -36,8 +36,9 @@ async function fetchLearnerPortalLinks(apiClient, userId) {
           // of the response so check if it exists before referencing fields
           branding_configuration: brandingConfiguration ? {
             logo: brandingConfiguration.logo || null,
-            banner_border_color: brandingConfiguration.banner_border_color || null,
-            banner_background_color: brandingConfiguration.banner_background_color || null,
+            primary_color: brandingConfiguration.primary_color || null,
+            secondary_color: brandingConfiguration.secondary_color || null,
+            tertiary_color: brandingConfiguration.tertiary_color || null,
           } : null,
           title: enterpriseCustomer.name,
           url: `${window.location.protocol}//${learnerPortalHostname}/${enterpriseCustomerSlug}`,


src/learnerPortalLinks.js
```diff
@@ -36,8 +36,8 @@ async function fetchLearnerPortalLinks(apiClient, userId) {
           // of the response so check if it exists before referencing fields
           branding_configuration: brandingConfiguration ? {
             logo: brandingConfiguration.logo || null,
-            banner_border_color: brandingConfiguration.banner_border_color || null,
-            banner_background_color: brandingConfiguration.banner_background_color || null,
+            banner_border_color: brandingConfiguration.secondary_color || null,
```
adamstankiewicz, line 39: Should we keep `banner_border_color` and `banner_background_color` or rename these to reflect the new names? Also, should tertiary color be included here as well to be complete?

Afaik, these values are not currently read anywhere.

kellmarie: Another question: would these primary and secondary colors be used in places other than the border and background colors here? I think having them be inconsistent here is fine if these primary and secondary color values are going to be used elsewhere on things that aren't explicitly a border or background color

iloveagent57: @jlajoie I think you have some open PRs that will make use of these fields, right?

adamstankiewicz: @iloveagent57 When I asked @jlajoie this morning if he was using these fields, he mentioned he was not and brought in the color fields just in case they were needed.

iloveagent57: OK!  I'll update this to a more reasonable thing, then. :)

iloveagent57: Updated to only include primary/secondary/tertiary.  If the values aren't read anywhere, let's just make it consistent with everything else now, as you said above.

src/learnerPortalLinks.js
```diff
@@ -36,8 +36,10 @@ async function fetchLearnerPortalLinks(apiClient, userId) {
           // of the response so check if it exists before referencing fields
           branding_configuration: brandingConfiguration ? {
             logo: brandingConfiguration.logo || null,
-            banner_border_color: brandingConfiguration.banner_border_color || null,
-            banner_background_color: brandingConfiguration.banner_background_color || null,
+            primary_color: brandingConfiguration.primary_color || null,
+            secondary_color: brandingConfiguration.secondary_color || null,
+            tertiary_color: brandingConfiguration.tertiary_color || null,
+
```
adamstankiewicz, line 42: super tiny nit: extra line

PR #16, opened 2020-08-11
fix: adds optional params to functions when process.env isn't configured

jlajoie:


src/service.js
```diff
@@ -1,4 +1,4 @@
-const fetchEnterpriseCustomers = apiClient => apiClient.get(`${process.env.LMS_BASE_URL}/enterprise/api/v1/enterprise-customer/`);
```
binodpant, line 1: can this defaulting be only done in the function above or is also needed here? ideally we could read this in one place defaulted to that env var.. but this is not a big issue, just that we have two places to change in case we use a different env var or sth in the future

binodpant:
Since it was an outage-causing type of issue, I would be more comfortable having this piece be captured in a unit test, a bit of extra time now hopefully will save much in future in case someone decides to change the env var scheme or any other assumptions. Would it be possible to just tight-test that part of the code which resolves the var without making this is a huge pull?

binodpant:
> Since it was an outage-causing type of issue, I would be more comfortable having this piece be captured in a unit test, a bit of extra time now hopefully will save much in future in case someone decides to change the env var scheme or any other assumptions. Would it be possible to just tight-test that part of the code which resolves the urls or the values of lmsBaseUrl etc without making this is a huge pull?

I am myself not a 100% on the effort-return on this one since it's such a small change, but leave it upto your judgement on whether a unit test is appropriate here, so consider this just a suggestion/idea-gargling

src/service.js
jlajoie: So the reason I left this here is because this function is importable. There's a scenario where you could `import { fetchEnterpriseCustomers } from 'frontend-enterprise/service'` and would then have the option of passing along a var or not depending on if it'll be set.

jlajoie:
> > Since it was an outage-causing type of issue, I would be more comfortable having this piece be captured in a unit test, a bit of extra time now hopefully will save much in future in case someone decides to change the env var scheme or any other assumptions. Would it be possible to just tight-test that part of the code which resolves the urls or the values of lmsBaseUrl etc without making this is a huge pull?
> 
> I am myself not a 100% on the effort-return on this one since it's such a small change, but leave it upto your judgement on whether a unit test is appropriate here, so consider this just a suggestion/idea-gargling

This whole repo needs tests sadly, though most of the issues I've been seeing aren't with logic and throwing errors, but with configuration so I might just punt on that. 

If it helps, my understanding of the outage was from the ``` https://courses.edx.org/enterprise/api/v1/enterprise-learner/ ``` endpoint and not the one we're hitting here. This fix is so this code will at least be executed since it currently isn't. If it helps alleviate concerns, we also have additional caching in place upstream of some of these calls that'll prevent the `enterprise-customer` endpoint from being spammed like what happened in the outage this weekend.

src/service.js
binodpant: I see, that makes sense... I can see a future mode where we have a function whose sole job is to resolve these values and both functions use it, and then we test that function. But it would not be a value investment for this stage of work I think, so all set

binodpant:
I see, thanks for the context there, and lol : `This whole repo needs tests sadly!` that seems like something to look into. All set for now!

PR #17, opened 2020-08-11
feat: expose useEnterpriseConfig hook

iloveagent57:
Several of our frontend apps and components require the same enterprise configuration data to fetch a user's learner portal link and make a decision about which dropdown items to show in the header menu:
* https://github.com/edx/frontend-component-header-edx/blob/master/src/data/hooks/enterprise.js
* https://github.com/edx/frontend-app-learning/pull/175

This PR exposes a new enterprise hook that can be used by the above.  It helps us DRY up the code and helps de-couple both of the above from the enterprise ecosystem.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,9 @@
   "publishConfig": {
     "access": "public"
   },
+  "watch": {
+    "build": "src/*.js"
+  },
   "scripts": {
     "build": "NODE_ENV=production BABEL_ENV=production webpack --config webpack.config.js",
     "gc": "commit",
@@ -17,7 +20,8 @@
     "prepublishOnly": "npm run build",
     "semantic-release": "semantic-release",
     "test": "jest --coverage",
-    "travis-deploy-once": "travis-deploy-once"
+    "travis-deploy-once": "travis-deploy-once",
+    "watch": "npm-watch build"
   },
   "repository": {
     "type": "git",
@@ -34,6 +38,7 @@
     "@commitlint/config-angular": "^6.0.2",
     "@commitlint/prompt": "^6.0.2",
     "@commitlint/prompt-cli": "^6.0.2",
+    "@edx/frontend-platform": "^1.5.2",
     "axios-mock-adapter": "^1.15.0",
     "babel-cli": "^6.26.0",
     "babel-core": "^6.26.3",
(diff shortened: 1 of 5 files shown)


src/hooks.js
```diff
@@ -0,0 +1,41 @@
+import getLearnerPortalLinks from './learnerPortalLinks';
+import { getSelectedEnterpriseUUID } from './utils';
+
+export default function useEnterpriseConfig(
+  authenticatedUser,
+  getAuthenticatedHttpClient,
+  useEffect,
+  useState,
```
iloveagent57, line 8: I ran into "Invalid hook call" issues trying to install react into this package and then install this package into `frontend-app-learner` https://reactjs.org/docs/error-decoder.html/?invariant=321 ...I think because react-dom had issues determining which react to use (the one in this package, or the one in frontend-app-learning).  Injecting the `useState` and `useEffect` functions from the client of this package fixed those up.

I ran into an error "no property `getAuthenticatedHttpClient()` of undefined" when I had frontend-platform installed, so I took the same injection approach with that, too.

src/hooks.js
```diff
@@ -0,0 +1,50 @@
...
+          learnerPortalLink.uuid === preferredUUID);
+        if (preferredLearnerPortalLink) {
+          const config = {
+            logoAltText: preferredLearnerPortalLink.title,
+            logoDestination: preferredLearnerPortalLink.url,
+          };
+          if (preferredLearnerPortalLink.branding_configuration) {
+            config.logo = preferredLearnerPortalLink.branding_configuration.logo;
+          }
+          setEnterpriseCustomerBrandingConfig(config);
```
adamstankiewicz, line 33: If `branding_configuration` is not set, I'm not sure we want/need to call `setEnterpriseCustomerBrandingConfig(config)` since a branding configuration does not exist.

package.json
```diff
@@ -64,5 +70,12 @@
       "!**/node_modules/**",
       "!**/tests/**"
     ]
+  },
+  "dependencies": {
+    "npm-watch": "^0.6.0"
```
adamstankiewicz, line 75: Should this be a `devDependency`?

src/hooks.js
iloveagent57: I wonder if `setEnterpriseCustomerBrandingConfig` is a bit of a misnomer - `config`, here, has the title and the url of the learner portal, and possibly the logo from the `branding_configuration`.  Maybe we can include the title and the logo (if it exists) in the `enterpriseLearnerPortalLink` state object and deprecate the `enterpriseCustomerBrandingConfig` object, since they're kinda doing the same thing?

iloveagent57:
I tried switching the build output filename to `index.js`.  Over in frontend-app-learning, when I changed the local module dist to `dist`, I encountered the multiple Reacts error again.  Switching back to `src` resolved it.  So I'm going to leave the whole headache alone for now.

PR #20, opened 2020-09-15
fix: only fetch user's preferred enterprise metadata

adamstankiewicz:
Pass the user's "preferred" enterprise uuid to the API call to `/enterprise-customer/` to prevent a paginated response.

**Testing instructions**
1. Install this branch of your local checkout of @edx/frontend-enterprise into a consuming MFE (e.g., frontend-app-learning).
2. Observe a network request to `/enterprise-customer/` which includes the `uuid` query parameter.

### DIFF
--- a/src/service.js
+++ b/src/service.js
@@ -1,4 +1,19 @@
-const fetchEnterpriseCustomers = (apiClient, lmsBaseUrl = process.env.LMS_BASE_URL) => apiClient.get(`${lmsBaseUrl}/enterprise/api/v1/enterprise-customer/`);
+import qs from 'querystring';
+
+const fetchEnterpriseCustomers = (
+  apiClient,
+  lmsBaseUrl = process.env.LMS_BASE_URL,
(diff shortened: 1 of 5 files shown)


src/service.js
```diff
@@ -1,4 +1,15 @@
-const fetchEnterpriseCustomers = (apiClient, lmsBaseUrl = process.env.LMS_BASE_URL) => apiClient.get(`${lmsBaseUrl}/enterprise/api/v1/enterprise-customer/`);
+import qs from 'querystring';
+
+const fetchEnterpriseCustomers = (
```
iloveagent57, line 3: Nit: `fetchPreferredEnterpriseCustomer`?  Also, since this will only ever return 0 or 1 things, should we have this do the `.pop()`?

iloveagent57 (commented):
👍 Looks good, just the one nit.

chavesj (approved):
LGTM

PR #21, opened 2020-11-12
ENT-3566: Refactor search common components out from learner portal so admin portal can also use it

binodpant:
Bringing in all course search related code under src/course-search here

in draft until following is fixed:

 - [X] npm run lint fails due to an error in no-cycles rule configuration https://edx-internal.slack.com/archives/CD4E4BCM9/p1605214473237100
 - [X] es5 check fails in vendor bundle
 - [X] need to use this dep locally in a cleaned up learner portal and ensure it works


src/course-search/data/utils.js
```diff
@@ -0,0 +1,15 @@
+export const sortItemsByLabelAsc = items =>
```
binodpant, line 1: moving this file and test here since no other code in LP uses it right now

package.json
```diff
@@ -38,25 +38,27 @@
...
     "axios-mock-adapter": "^1.15.0",
-    "babel-cli": "^6.26.0",
-    "babel-core": "^6.26.3",
-    "babel-loader": "^7.1.5",
-    "babel-plugin-rewire": "^1.2.0",
-    "babel-plugin-transform-object-rest-spread": "^6.26.0",
-    "babel-plugin-transform-runtime": "^6.23.0",
-    "babel-preset-env": "^1.7.0",
-    "babel-preset-react": "^6.24.1",
-    "babel-runtime": "^6.26.0",
```
binodpant, line 51: these deps can go due to frontend-build

public/index.html
```diff
@@ -0,0 +1,11 @@
+<!doctype html>
```
binodpant, line 1: this was needed by the fedx-scripts webpack command

README.rst
```diff
@@ -5,6 +5,16 @@ frontend-enterprise
 
 frontend-enterprise contains utility functions for supporting enterprise features.
 
+Dependency notes
+-----
+
+v6 and higher of query-string will fail the es5 check
+eslint-plugin-import needed at least 2.22.1 to avoid failures with an infinity symbol (see https://stackoverflow.com/questions/64790681/eslint-error-configuration-for-rule-import-no-cycle-is-invalid)
```
binodpant, line 12: I will also likely update this version in frontend-build, ran this idea past @davidjoy  already but having the dep in our package.json unblocks this PR

.eslintrc.js
```diff
@@ -0,0 +1,20 @@
+const { getBaseConfig } = require("@edx/frontend-build");
+
+const config = getBaseConfig("eslint");
+
+config.rules = {
+  "import/no-extraneous-dependencies": [
```
raq929, line 6: I'm assuming this stuff makes it the same as `frontend-app-learner-portal-enterprise`? 
I wonder if we should use this file as a joint `.eslintrc` for all our enterprise repos so they are all on the same version of everything....

package.json
```diff
@@ -10,16 +10,16 @@
     "build": "src/*.js"
   },
   "scripts": {
-    "build": "NODE_ENV=production BABEL_ENV=production webpack --config webpack.config.js",
+    "build": "fedx-scripts webpack",
```
raq929, line 13: Yessss seems great

package.json
```diff
@@ -73,10 +77,24 @@
...
+    "react-redux": "^7.2.1",
+    "react-router-dom": "^5.2.0",
+    "react-loading-skeleton": "^2.1.1",
+    "redux": "^4.0.0"
   },
   "dependencies": {
-    "query-string": "^6.13.2"
+    "classnames": "^2.2.5",
+    "query-string": "^5.1.1",
+    "@fortawesome/fontawesome-svg-core": "^1.2.17",
```
raq929, line 92: hahahaha we are putting them in only to remove them later with the rebrand. 😢 
(No issues, but I'll make a rebrand ticket for this repo now.)

src/hooks.js
```diff
@@ -1,42 +1,83 @@
...
+
+export const useIsFirstRender = () => {
+  const isMountRef = useRef(true);
+  useEffect(() => {
+    isMountRef.current = false;
+  }, []);
+  return isMountRef.current;
+};
+
+export default function useEnterpriseConfig(
```
raq929, line 40: I would actually change this to not be a default export anymore (you'll have to disable an eslint rule for the file; that's expected in hooks files). You'll have to change the imports to import `{ useEnterpriseConfig }`.

src/utils/tests.jsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import { Router } from 'react-router-dom';
+import { createMemoryHistory } from 'history';
+import { render } from '@testing-library/react';
+import { AppContext } from '@edx/frontend-platform/react';
```
raq929, line 5: Hmmmm.... this is going to be interesting when you port it to the admin portal. Presumably it has a different AppContext. We might want to rename and / or check what the admin portal does, because I think `AppContext` should probably live in it's app, not in the shared repo. If they're similar enough I could be wrong, though.

raq929 (approved):
One small comment about your hooks file, and I thing you should take a look at the AdminPortal AppContext (if it exists, or find where it's getting the enterprise data that's used for search) before we merge, but otherwise, it's looking good!

src/utils/tests.jsx
binodpant: good point!

src/hooks.js
binodpant: good point!

src/utils/tests.jsx
binodpant: hmm @raq929 I don't see it being used at all! and now have this in both repos after refactor. Perhaps we can remove it..? I also did not find any usage of AppContext in admin portal repo

.eslintrc.js
binodpant: interesting thought!

src/utils/tests.jsx
binodpant: will leave it here for now, removing from learner portal, the same function for now

README.rst
adamstankiewicz: @binodpant I can look into upgrading eslint-plugin-import in frontend-build, or review any PRs for it. I'm overdue to look at some of the upgrade PRs for frontend-build as the primary technical "owner" :)

package.json
```diff
@@ -10,16 +10,16 @@
...
   },
   "scripts": {
-    "build": "NODE_ENV=production BABEL_ENV=production webpack --config webpack.config.js",
+    "build": "fedx-scripts webpack",
     "gc": "commit",
     "commitmsg": "commitlint -e $GIT_PARAMS",
     "coveralls": "cat ./coverage/lcov.info | coveralls",
-    "lint": "eslint --ext .js --ext .jsx .",
+    "lint": "fedx-scripts eslint --ext .js --ext .jsx .",
     "is-es5": "es-check es5 ./dist/*.js",
```
adamstankiewicz, line 18: Should we make frontend-enterprise consistent with our other edX libraries (e.g., Paragon, frontend-platform), which don't have the is-es5 check as a script or called in .travis.yml?

package.json
adamstankiewicz: Thanks @raq929!

src/utils/tests.jsx
adamstankiewicz: Admin portal does not use frontend-build or frontend-platform right now, so it does not utilize AppContext.

adamstankiewicz (approved):
:shipit:

package.json
binodpant: yeah makes sense if the MFEs do it. George will remove travis.yml file as well

PR #23, opened 2020-11-18
feat: use prod webpack base config with overrides

binodpant:
In the prior PR https://github.com/edx/frontend-enterprise/pull/21 I used the base webpack.prod.config.js from frontend-build which is tuned for applications not libraries, hence it generated a number of files via webpack that change the output of the library 

hence importing failed in frontend-app-learner-portal-enterprise when I used the latest release of `@edx/frontend-enterprise`

The desired assets were 

```
index.js
index.js.map
```

instead we were getting a whole lot of files including index.html and other chunked files.

This change uses the base config for webpack-prod preset but overrides settings to produce index.js and index.js.map like it used to, before my change in the prior PR


raq929 (approved):

webpack.prod.config.js
```diff
@@ -0,0 +1,33 @@
+const path = require('path');
+const UglifyJsPlugin = require('uglifyjs-webpack-plugin');
+const { getBaseConfig } = require('@edx/frontend-build');
+
+const origConfig = getBaseConfig('webpack-prod');
+
+const config = {
+  ...origConfig,
```
binodpant, line 8: I use the orig config from base frontend-build in case enhancements are needed, but I still largely use the rest of the config as it was before, I did not research into whether I could skip the optimizations override, I just wanted to get one file for now in the output but we could research into using the chunks for index.js later

PR #25, opened 2020-11-18
feat: ENT-3717 fix bundling no webpack use babel instead

binodpant:
This release fixes a lot of issues in the bundling I created in the prior release, and the output folder dist/ looks like this 

This brings this repo in line with other library repoes where the MFEs that use these do the webpack based optimizations instead of the libs doing it

```
ᐅ ls -al dist                                                     
total 224
drwxr-xr-x  19 binodpant  staff    608 Nov 19 17:07 .
drwxr-xr-x  20 binodpant  staff    640 Nov 19 17:30 ..
-rw-r--r--   1 binodpant  staff  34520 Nov 19 17:07 LICENSE
-rw-r--r--   1 binodpant  staff   1907 Nov 19 17:07 README.rst
-rw-r--r--   1 binodpant  staff    435 Nov 19 17:07 constants.js
-rw-r--r--   1 binodpant  staff    889 Nov 19 17:07 constants.js.map
drwxr-xr-x  29 binodpant  staff    928 Nov 19 17:07 course-search
-rw-r--r--   1 binodpant  staff   4657 Nov 19 17:07 hooks.js
-rw-r--r--   1 binodpant  staff   5241 Nov 19 17:07 hooks.js.map
-rw-r--r--   1 binodpant  staff    761 Nov 19 17:07 index.js
-rw-r--r--   1 binodpant  staff   1574 Nov 19 17:07 index.js.map
-rw-r--r--   1 binodpant  staff   6753 Nov 19 17:07 learnerPortalLinks.js
-rw-r--r--   1 binodpant  staff   7707 Nov 19 17:07 learnerPortalLinks.js.map
-rw-r--r--   1 binodpant  staff   2932 Nov 19 17:07 package.json
-rw-r--r--   1 binodpant  staff    734 Nov 19 17:07 service.js
-rw-r--r--   1 binodpant  staff   1240 Nov 19 17:07 service.js.map
drwxr-xr-x   4 binodpant  staff    128 Nov 19 17:07 utils
-rw-r--r--   1 binodpant  staff   2015 Nov 19 17:07 utils.js
-rw-r--r--   1 binodpant  staff   4281 Nov 19 17:07 utils.js.map
```

tested locally with learner portal


adamstankiewicz (approved):
LGTM! 🔥

package.json
```diff
@@ -10,12 +10,11 @@
     "build": "src/*.js"
   },
   "scripts": {
-    "build": "fedx-scripts webpack",
+    "build": "make build",
```
raq929, line 13: Is there not a fedex-script that does what we want? Is this kind of build something we would want to move into fedx-scripts?

raq929 (approved):

package.json
binodpant: good point yeah. I think we would have to put some thought there indeed. I find that the published recipe there is only designed for apps, hence I had to discover stuff like this by looking at frontend-platform etc. I think there is room for us to publish library friendly recipe in frontend-build... or at least a doc recipe if not code recipe

This model was from frontend-platform but I agree a fedx-scripts library-build or something would be better

PR #27, opened 2020-11-19
Move travis ci & semantic release steps to github actions

georgebabey:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,20 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2  
+    - name: Setup Nodejs
+      uses: actions/setup-node@v1
+      with:
+        node-version: 12
+    - name: Install dependencies
+      run: npm ci
```
binodpant, line 16: I like that we use npm ci instead of npm i !!!

.github/workflows/release.yml
```diff
@@ -0,0 +1,25 @@
...
+          node-version: 12
+      - name: Install dependencies
+        run: npm ci
+      - name: Build
+        run: npm run build
+      - name: Release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.SEMANTIC_RELEASE_NPM_TOKEN }}
+        run: npm run semantic-release
```
binodpant, line 25: have not looked at how we use coveralls, don't need it anymore?

binodpant (approved):
left minor comments, but nothing I see to not merge this

.github/workflows/release.yml
georgebabey: Good call - When I created this PR, it wasn't wired up for this repo so I didn't include it. but it looks like @adamstankiewicz  did add it so let me add that back

PR #61, opened 2021-02-18
fix: add styles to SearchHeader

raq929:
Adds missing styles to the SearchHeader
Adds a build process to create a css file (stolen with modifications  from paragon)


package.json
```diff
@@ -85,6 +84,7 @@
     "@fortawesome/free-solid-svg-icons": "^5.8.1",
     "@fortawesome/react-fontawesome": "^0.1.4",
     "classnames": "^2.2.5",
+    "node-sass": "^5.0.0",
```
adamstankiewicz, line 87: node-sass is actually now deprecated in favor of using Dart sass.

image

We updated frontend-build to use Dart sass a few months back: See https://github.com/edx/frontend-build/pull/113 and https://github.com/edx/frontend-build/issues/111.

We'll likely want to make a similar change to Paragon.

edx-abolger: Per conversation Lael and I had at standup: 1. If Dart sass works, great 2. If Dart sass doesnt work out of box for this, I am strongly in favor of webpack (even if it seems overkill) compared to any completely-new-to-edX alternatives because I do not want us to have 1 repo using a compiler technology: that introduces tooling debt at the organizational level.

raq929: Dart-sass did work, and I brought the build issue up at the fedex meeting and David Joy alerted me to a different webpack config that's available via frontend-build that we can use to build this. Gonna open a ticket for that.

Makefile
```diff
@@ -6,3 +6,4 @@ build:
 	cp ./package.json ./dist/package.json
 	cp ./LICENSE ./dist/LICENSE
 	cp ./README.rst ./dist/README.rst
+	node build-scss.js
```
edx-abolger, line 9: Will this get you the correct version of node anywhere this makefile is supposed to run?

edx-abolger (approved):
I approve but also we need a version bump and change log (that can be a separate PR if it makes more sense)

Makefile
raq929: Yeah, the node version is set in the `.nvmrc` file.

binodpant: The github actions hardcodes a node version right now, I think we have not made is support .nvmrc yet. but locally with `nvm use`, yes.

PR #68, opened 2021-02-26
Update build process and make filters configurable

raq929:
Compiles .scss using `babel-preserve-modules` instead of doing it seperately
Moves dependencies to peer / dev dependencies to avoid conflicts with other packages
Updates filters to make them configurable.


package.json
```diff
@@ -82,14 +88,15 @@
...
     "@fortawesome/free-solid-svg-icons": "^5.8.1",
     "@fortawesome/react-fontawesome": "^0.1.4",
+    "algoliasearch": "^4.8.5",
     "classnames": "^2.2.5",
     "query-string": "^5.1.1",
     "react-instantsearch-dom": "^6.8.3",
     "react-loading-skeleton": "^2.1.1"
-  }
+  },
+  "dependencies": {}
```
binodpant, line 101: is an empty list needed?

package.json
```diff
@@ -82,14 +88,15 @@
...
     "react-redux": "^7.2.1",
     "react-router-dom": "^5.2.0",
-    "redux": "^4.0.0"
-  },
-  "dependencies": {
+    "react-test-renderer": "17.0.1",
+    "redux": "4.0.5",
     "@fortawesome/free-solid-svg-icons": "^5.8.1",
     "@fortawesome/react-fontawesome": "^0.1.4",
+    "algoliasearch": "^4.8.5",
```
binodpant, line 95: I lost track of slack, I will try to dig it up there was an issue when I had introduced some peer deps, in build process of some other MFEs that use this project (I was surprised to learn than even ecommerce builds somehow depend on this, let me check and get back to you on this) @adamstankiewicz may also be able to add context

binodpant: slack thread for relevant convos: https://edx-internal.slack.com/archives/CM4RPEYEQ/p1606157355298600

adamstankiewicz: frontend-enterprise is used by [frontend-component-header-edx](https://github.com/edx/frontend-component-header-edx/blob/55634492ac2b300467d39d0c6ed11be4b0760c7b/src/Header.jsx#L13) to fetch whether a learner is linked to an Enterprise with the learner portal enabled, and if so changes the logo image/link and the avatar dropdown menu items to point to https://enterprise.edx.org/{slug} instead of the b2c links.

This header component is used by various MFEs, hence the dependency by some of our apps. Similar story with the frontend-enterprise dependency in frontend-app-learning, which has its own header.

I'm not recalling the particular issue @binodpant is referring to, though, unfortunately.

package.json
```diff
@@ -82,14 +88,15 @@
     "react-dom": "^17.0.1",
     "react-redux": "^7.2.1",
     "react-router-dom": "^5.2.0",
-    "redux": "^4.0.0"
-  },
-  "dependencies": {
+    "react-test-renderer": "17.0.1",
```
adamstankiewicz, line 91: Should this be a dependency? If it's only intended to run during local development, I believe this should be a `devDependency`?

binodpant: +1 looks like a devDep

binodpant:
suggesting making this a new major version, barring that I am good with this!

binodpant (approved):
:shipit:

package.json
raq929: I'm not sure it's needed, but I feel like, in the absence of being able to have comments in JSON, explicitly leaving it in lets everyone know that this was a deliberate choice.

package.json
raq929: We're making this a major version bump, so it won't get automatically pulled in, and people can add the dependencies need to satisfy the peerDependencies.

raq929:
@davidjoy You'll only need algolia-related dependencies and `react-loading-skeleton` if you're using the algolia related components. That's actually been true for a while; and I think this move to make them peerDependencies instead of bundled in I think will be good for the MFEs that aren't using those components.

PR #71, opened 2021-03-01
Remove unneeded hook

raq929:
I have moved the hook to the learner portal, as it is only relevant there; I am exporting it's dependencies from this library


binodpant:
I like this! if the hook has been specific to learner portal use case it should be there. we can revisit the design of course-search as a component to be reused, and this seems like a correct next move. data fetching and index-interaction with algolia live in MFE and only the core component logic and base styles markup here in enterprise, good!

binodpant (approved):
lgtm!

PR #72, opened 2021-03-01
Add tests and run coveralls on PRs

raq929:
Adds tests to hopefully fix the drop in % test coverage that happened when I removed the hook.
Has coveralls run on PRs, because if it's going to fail the build, I want to know before merge.


binodpant (approved):
nice! +1 for the coveralls addition

PR #79, opened 2021-03-23
feat: add language facet to SearchHeader

adamstankiewicz:
[ENT-2990](https://openedx.atlassian.net/browse/ENT-2990)

Adds a "Language" facet dropdown to the course search header.

image

### DIFF
--- a/src/course-search/SearchContext.jsx
+++ b/src/course-search/SearchContext.jsx
@@ -53,11 +53,14 @@ const SearchData = ({ children, searchFacetFilters }) => {
     }
   }, [newQueryString]);
 
-  const value = {
-    refinementsFromQueryParams,
-    dispatch,
-    searchFacetFilters,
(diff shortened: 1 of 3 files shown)


src/course-search/SearchContext.jsx
```diff
@@ -53,11 +53,14 @@ const SearchData = ({ children, searchFacetFilters }) => {
     }
   }, [newQueryString]);
 
-  const value = {
-    refinementsFromQueryParams,
-    dispatch,
-    searchFacetFilters,
-  };
+  const value = useMemo(
```
adamstankiewicz, line 56: Note: Context values should always be memoized to avoid unnecessary re-rendering of children components that rely on the context values.

kellmarie (approved):

iloveagent57 (approved):

PR #84, opened 2021-03-29
fix: add style variants

raq929:
So that the header can work on a a white or dark background

"Default" (which is not actually the default, but does use default styles for the components. If you've got a better name, LMK)
Screen Shot 2021-03-29 at 12 29 32 PM

"Inverse" Meant for dark backgrounds. Used as the default because our current usage is all dark backgrounds.
Screen Shot 2021-03-29 at 12 37 24 PM


raq929:
The weird corners in the top photo is a Paragon thing; I'm working on it separately.

src/course-search/TypeaheadFacetDropdown.jsx
```diff
@@ -43,6 +44,7 @@ const TypeaheadFacetDropdown = ({
       title={title}
       isBold={isBold}
       type="typeahead"
+      {...props}
```
edx-abolger, line 47: For my own learning: it feels like spread operator with {} around it should cancel each other out. Are you doing this as a cheap way of making an immutable copy of this data?

src/course-search/styles/_FacetList.scss
```diff
@@ -28,13 +28,11 @@ $facet-dropdown-max-height: 250px;
...
     }
   }
-
   .typeahead {
     .dropdown-menu {
       max-height: unset;
       overflow-y: unset;
-
-      .typeahead-dropdown-menu-scrollable-items {
+          .typeahead-dropdown-menu-scrollable-items {
```
edx-abolger, line 35: Nit: is this the accepted indentation?

src/course-search/styles/_SearchField.scss
```diff
@@ -0,0 +1,14 @@
...
+.fe__searchfield-input-box {
+  @extend .mb-2;
+  @extend .text-brand-primary !optional;
+  @extend .font-weight-normal;
+}
+
+.fe__searchfield--inverse {
+  border: 0 !important;
+  @extend .bg-white;
+}
```
binodpant, line 14: newline

src/course-search/styles/_SearchBox.scss
```diff
@@ -0,0 +1,7 @@
+.fe__searchbox-col--default {
+  padding: 0;
+
+  .container, .row {
+    padding: 0;
+  }
+}
```
binodpant, line 7: newline

src/course-search/styles/_SearchField.scss
```diff
@@ -0,0 +1,14 @@
...
+
+.fe__searchfield-input-box {
+  @extend .mb-2;
+  @extend .text-brand-primary !optional;
+  @extend .font-weight-normal;
+}
+
+.fe__searchfield--inverse {
+  border: 0 !important;
```
edx-abolger, line 12: What's the important flag overriding?

edx-abolger (approved):

binodpant:
looks awesome!

binodpant (approved):
:shipit:

src/course-search/styles/_SearchField.scss
raq929: Bootstrap almost always throws `!important` on their css classes, which annoys me greatly, because then it's hard to get enough specificity to override them. But, that's what this is doing, overriding a bootstrap class.

src/course-search/styles/_FacetList.scss
raq929: No, it's a mistake, thanks.

src/course-search/TypeaheadFacetDropdown.jsx
raq929: No, it's inside a JSX tag, so the spread operator here will take whatever is in props and add them as props to this component. So, if what's in props is `{ color: 'blue', variant: 'yellow'}` this would be the same as writing
```
color='blue'
variant='yellow'
``` 
on these lines.

I suspect the syntactic sugar of JSX is turning the everything above this into an object somewhere under the hood, anyway, and this is just extending that object. 
It's... not the choice I would have made had I written React. But here we are.

PR #87, opened 2021-04-09
fix: update search facet filters

raq929:
Updates the search context to correctly set query params on additional filters it is given.
Updates paragon and makes the required style changes
Exports all available actions that might be needed within the SearchContext


src/course-search/SearchFilters.jsx
```diff
@@ -62,6 +62,7 @@ const SearchFilters = ({ variant }) => {
           typeaheadOptions={typeaheadOptions}
           searchable={!!typeaheadOptions}
           variant={variant}
+          noDisplay={noDisplay}
```
alex-sheehan-edx, line 65: 🎉

alex-sheehan-edx (approved):
✅

PR #91, opened 2021-04-15
fix: clean up facet pill styles

adamstankiewicz:
## Current

image

## Updated
image

Note the pill is 4px shorter. Also, both the checkbox and pill are top-aligned to the facet item, rather than awkwardly being centered.

### DIFF
--- a/src/course-search/SearchBox.jsx
+++ b/src/course-search/SearchBox.jsx
@@ -40,7 +40,7 @@ export const SearchBoxBase = ({
   return (
     <div className={className}>
       {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
-      <label id="search-input-box" className="fe__searchfield-input-box">
+      <label id="search-input-box" className="fe__searchfield-input-box text-brand-primary">
         {searchText}
       </label>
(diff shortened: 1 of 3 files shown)


src/course-search/SearchBox.jsx
```diff
@@ -40,7 +40,7 @@ export const SearchBoxBase = ({
   return (
     <div className={className}>
       {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
-      <label id="search-input-box" className="fe__searchfield-input-box">
+      <label id="search-input-box" className="fe__searchfield-input-box text-brand-primary">
```
adamstankiewicz, line 43: This will fix the issue where the "Search courses" label on the search box isn't turning white on darker backgrounds.

iloveagent57 (approved):

src/course-search/SearchBox.jsx
raq929: Thanks, sorry I didn't get to it!

adamstankiewicz: No worries!

PR #92, opened 2021-04-22
fix: replacing query string parse with built in functionality

alex-sheehan-edx:
In order to appease `is-es5`, we need to replace the functionality of `query-string`'s `parse` method with the compatible `URLSearchParams` interface. In order to perfectly match behaviors, I've added a util function to convert the `URLSearchParams` object to a native dictionary.

### DIFF
--- a/src/course-search/config/index.js
+++ b/src/course-search/config/index.js
@@ -1,15 +1,9 @@
-import qs from 'query-string';
-
 import {
   FEATURE_ENROLL_WITH_CODES,
   FEATURE_LANGUAGE_FACET,
   FEATURE_PROGRAM_TITLES_FACET,
 } from '../../constants';
-
-const hasFeatureFlagEnabled = (featureFlag) => {
-  const { features } = qs.parse(window.location.search);
-  return features && features.split(',').includes(featureFlag);
-};
+import { hasFeatureFlagEnabled } from '../data/utils';
 
 const features = {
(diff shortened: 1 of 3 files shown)


edx-abolger (approved):

src/course-search/config/index.js
```diff
@@ -1,13 +1,13 @@
-import qs from 'query-string';
-
 import {
   FEATURE_ENROLL_WITH_CODES,
   FEATURE_LANGUAGE_FACET,
   FEATURE_PROGRAM_TITLES_FACET,
 } from '../../constants';
+import { paramsToObject } from '../data/utils';
```
raq929, line 6: I think this function exists in `@edx/frontend-platform` and we should import it from there?

raq929 (approved):
One thing we should possibly import from frontend-platform, otherwise, looking good!

src/course-search/config/index.js
```diff
@@ -1,13 +1,13 @@
...
 
 const hasFeatureFlagEnabled = (featureFlag) => {
```
binodpant, line 8: esp. given we are changing how it's implemented, worth moving `hasFeatureFlagEnabled` into a util function and having a test I would suggest, it's kind of an important function and will help us avoid having to manually test this

src/course-search/config/index.js
alex-sheehan-edx: I believe that function has been deprecated ☹️

binodpant (approved):
lgtm! do check it out in staging before prod but this looks like it will work, try sth like `?features=BULK_ENROLLMENT` and also trying without feature on

PR #93, opened 2021-04-23
fix: replacing query-string with built in functionality

alex-sheehan-edx:
In an effort to allow filter work to be es5 compatible, this PR is removing another instance of `query-string` which has presumably been causing headaches with polyfills.

### DIFF
--- a/src/course-search/SearchContext.jsx
+++ b/src/course-search/SearchContext.jsx
@@ -3,11 +3,10 @@ import React, {
 } from 'react';
 import PropTypes from 'prop-types';
 import { useLocation, useHistory } from 'react-router-dom';
-import qs from 'query-string';
 import { BOOLEAN_FILTERS, SEARCH_FACET_FILTERS } from './data/constants';
 import { refinementsReducer } from './data/reducer';
 import { setMultipleRefinementsAction } from './data/actions';
-import { updateRefinementsFromQueryParams } from './data/utils';
+import { paramsToObject, stringifyRefinements, updateRefinementsFromQueryParams } from './data/utils';
 import { useIsFirstRender } from '../hooks';
 
 export const SearchContext = createContext();
@@ -37,8 +36,7 @@ const SearchData = ({ children, searchFacetFilters }) => {
   const { search } = useLocation();
   const history = useHistory();
 
-  const queryParams = useMemo(() => qs.parse(search), [search]);
-
+  const queryParams = useMemo(() => paramsToObject(new URLSearchParams(search)), [search]);
   useEffect(() => {
     const activeFacetAttributes = searchFacetFilters.map(filter => filter.attribute);
     const refinementsToSet = getRefinementsToSet(queryParams, activeFacetAttributes);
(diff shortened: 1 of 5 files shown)


raq929 (approved):

binodpant:
FYI 

https://github.com/edx/frontend-enterprise/search?q=query-string&type=code

looks like I had added a note a while ago .. don't need to strictly speaking but may also want to clear out that readme.rst line about query-string

alex-sheehan-edx:
I actually think with this PR we can fully remove query-string dependency! 🎉 Any objections? 👀

src/course-search/SearchContext.jsx
```diff
@@ -47,7 +45,12 @@ const SearchData = ({ children, searchFacetFilters }) => {
 
   const newQueryString = useMemo(() => {
     const refinementsWithJoinedLists = updateRefinementsFromQueryParams(refinementsFromQueryParams);
-    return qs.stringify(refinementsWithJoinedLists);
+    let refinementString = new URLSearchParams(refinementsWithJoinedLists);
+
+    if (refinementString) {
+      refinementString = refinementString.toString().replace(/[+]/g, '%20');
```
raq929, line 51: Talk to me about what this is doing? Also, I suspect that if we need to do it here, we probably need to do it for *all* query strings, and it should be moved to a function, probably it should be part of `refinementsWithJoinedLists`. My worry is that if we need to do this replacement, we likely need to do others as well, and that's when we should be looking for a library that does this.

raq929:
And yes, please remove query-string as a dependency as it is causing our builds to fail.

src/course-search/SearchContext.jsx
alex-sheehan-edx: testing locally this is what I found:
```
**refinements:** 
`{subjects: "Architecture,Computer Science,Humanities"}`

**query-string's stringify:**
`subjects=Architecture%2CComputer%20Science%2CHumanities`

**new URLSearchParams(refinements).toString()**
`subjects=Architecture%2CComputer+Science%2CHumanities`

 **new URLSearchParams(refinements).toString().replace(/[+]/g, '%20')**
 `subjects=Architecture%2CComputer%20Science%2CHumanities`
 
 The headache here was that `new URLSearchPArams().toString() wasn't encoding the `+` for spaces so I needed to manually encode them (since we also don't want to encode the `=`). I used regex because we can potentially have many and `replace('+', '%20')` only grabs the first instance.

Comment (alex-sheehan-edx) [reply src/course-search/SearchContext.jsx:51]: FWIW I think this is the only time we use `stringify` in frontend-enterprise so no need to replace it anywhere else.

Comment (alex-sheehan-edx) [reply src/course-search/SearchContext.jsx:51]: and the conditional 
```
if (refinementString) {
```
is there to make sure we're handling cases where this doesn't exist.

raq929: What happens if you pass URLSearchParams the params with the original arrays instead of the joined lists? It might do that joining for us.

raq929: Then we might be able to get rid of `refinementsWithJoinedLists` entirely.

alex-sheehan-edx: it seems like specifically the spaces within individual refinements ie `computer science` aren't being joined and those aren't addressed by the original array, sadly

raq929 (approved):

PR #95, opened 2021-04-26
feat: :sparkles: Tracking of search action ENT-4126

binodpant:
When search button is clicked or search initiated, send tracking event with query string data as the payload data

Tracking is only invoked if the `SearchData` context is populated with `trackingEvent` which only happens if one uses SearchData like this

```
<SearchData trackingName='public_enterprise_catalog'>...
```

when present, this is concatenated with other strings to form event name lke this 

```
edx.enterprise.<trackingName>.catalog_search.query_submitted
```

I put this in context instead of making it a prop since more than one component within  or next to SearchHeader could find use for this when doing further tracking

JIRA: ENT-4126

Note: I also started process of getting segment setup for prod and stage by requesting data engineering: https://openedx.atlassian.net/servicedesk/customer/portal/16/DESUPPORT-854?created=true as per https://openedx.atlassian.net/wiki/spaces/FEDX/pages/909738013/How+to+Create+and+Deploy+a+new+Microfrontend+inside+edX#Setup-Segment

Once that is done I will do a PR for adding segment keys for prod/stage

This PR is just to add the tracking to meet one of the ACs of ENT-4126 (search action tracking)


src/course-search/SearchBox.jsx
```diff
@@ -4,19 +4,25 @@ import classNames from 'classnames';
...
 import { deleteRefinementAction, setRefinementAction } from './data/actions';
 import { STYLE_VARIANTS } from '../constants';
 import { SearchContext } from './SearchContext';
 import { QUERY_PARAM_FOR_PAGE, QUERY_PARAM_FOR_SEARCH_QUERY } from './data/constants';
 
 export const searchText = 'Search courses';
+// this prefix will be combined with one of the SearchBox props to create a full tracking event name
+// only if event name prop is provided by user. In the absence of the tracking name prop,
+// no tracking event will be sent.
+export const SEARCH_EVENT_NAME_PREFIX = 'edx.enterprise';
```
raq929, line 18: I think we'll eventually want to add similar tracking to learner portal and admin portal, in which case, this prefix may not be specific enough.

src/course-search/tests/SearchBox.test.jsx
```diff
@@ -44,4 +50,17 @@ describe('<SearchBox />', () => {
     expect(history).toHaveLength(3);
     expect(history.location.search).toEqual('');
   });
+  test('tracking event when search initiated with trackingName present in context', () => {
+    renderWithSearchContextAndTracking(<SearchBoxBase />, 'aProduct');
+
+    // fill in search input and submit the search
+    fireEvent.change(screen.getByRole('searchbox'), { target: { value: TEST_QUERY } });
+    fireEvent.click(screen.getByText('submit search'));
```
raq929, line 58: The `react-testing-library` docs say we should use `userEffect` instead of `fireEvent`. Why they still provide `fireEvent` is a mystery to me, but let's keep up-to-date with their recommendations.

raq929 (approved):
A couple of comments, but it looks good to me!

src/course-search/SearchBox.jsx
adamstankiewicz: For learner/admin portals, couldn't we just change the `trackingName` prop passed to the `SearchData` component to include `learner_portal` and `admin_portal` respectively, resulting in `edx.enterprise.learner_portal.catalog_search` as the event name?

src/course-search/SearchBox.jsx
```diff
@@ -26,6 +32,11 @@ export const SearchBoxBase = ({
   const handleSubmit = (searchQuery) => {
     dispatch(setRefinementAction(QUERY_PARAM_FOR_SEARCH_QUERY, searchQuery));
     dispatch(deleteRefinementAction(QUERY_PARAM_FOR_PAGE));
+    if (trackingName) {
+      sendTrackEvent(`${SEARCH_EVENT_NAME_PREFIX}.${trackingName}.catalog_search`, {
```
adamstankiewicz, line 36: Might suggest renaming to ``${SEARCH_EVENT_NAME_PREFIX}.${trackingName}.catalog_search.query_submitted`` or similar so that we could expand on the `catalog_search` namespace with future events that aren't directly related to search submission.

src/course-search/tests/SearchBox.test.jsx
binodpant: ah yes indeed, https://testing-library.com/docs/ecosystem-user-event , done!

src/course-search/SearchBox.jsx
binodpant: yeah that was my vision, is for users to provide that middle portion of the name (then prefix and suffix would then just be added by frontend-enterprise)

src/course-search/SearchBox.jsx
binodpant: I see what you mean, I like it.. I can see `catalog_search.filter_applied` etc too

src/course-search/SearchBox.jsx
binodpant: yes so 

```
<SearchData trackingName="admin_portal">
```
kind of thing is what I am thinking. It just needs to be unique per product I guess.

PR #96, opened 2021-04-27
fix: disable side effects to not bundle all of the react testing library

alex-sheehan-edx:
`renderWithSearchContext` was being exported from `/utils/tests` but side effects were not disabled so our frontend apps were being bundled with all of the react test library

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,6 @@
 {
   "name": "@edx/frontend-enterprise",
+  "sideEffects": false,
   "version": "1.0.0-semantically-released",
   "description": "Frontend utilities for supporting enterprise features.",
   "main": "dist/",


davidjoy (approved):

raq929:
Wait, tell me more about this sideEffects thing and what it does?

alex-sheehan-edx:
> Wait, tell me more about this sideEffects thing and what it does?

I got a wonderful explanation of Tree Shaking from David Joy! (along with [this](https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free) great documentation) TL:DR Tree Shaking, as I understand it, is the term used to refer to eliminating dead code. 

Due to the way that we've built and distributed this library, tree shaking isn't possible or at least isn't done when it's built. So, by adding `"sideEffects": false` to the package json, we are informing webpack that our code is side effect-less and that it can safely prune the unused `testing-library` when building.

PR #97, opened 2021-04-30
refactor!: separate components into npm packages in monorepo

adamstankiewicz:
BREAKING CHANGE: refactor to split ui components and logic into separate packages using Lerna

The result of this change is that `@edx/frontend-enterprise` will no longer be published as its own NPM package. Instead, the package is split into 3 separate packages:

1. `@edx/frontend-enterprise-utils`
2. `@edx/frontend-enterprise-catalog-search`
3. `@edx/frontend-enterprise-logistration`

See the drafted ADR in this PR for more details: `docs/decisions/0002-lerna-monorepo.rst`

image

Note this does not remove/change any existing functionality of the code; just a lot of moving files / code around into separate packages. It does, however, add a component: `LoginRedirect` in the `@edx/frontend-enterprise-logistration` package.


README.rst
```diff
@@ -21,31 +47,14 @@ type must be one of [build, ci, docs, feat, fix, perf, refactor, revert, style,
...
-
-.. |Build Status| image:: https://api.travis-ci.com/edx/frontend-enterprise.svg?branch=master
-   :target: https://travis-ci.com/edx/frontend-enterprise
-.. |Coveralls| image:: https://img.shields.io/coveralls/edx/frontend-enterprise.svg?branch=master
-   :target: https://coveralls.io/github/edx/frontend-enterprise
-.. |npm_version| image:: https://img.shields.io/npm/v/@edx/frontend-enterprise.svg
-   :target: @edx/frontend-enterprise
-.. |npm_downloads| image:: https://img.shields.io/npm/dt/@edx/frontend-enterprise.svg
-   :target: @edx/frontend-enterprise
-.. |license| image:: https://img.shields.io/npm/l/@edx/frontend-enterprise.svg
```
raq929, line 48: We should probably still be linking to the license in the readme.  Though I'm not sure this ever actually did.

docs/decisions/0002-lerna-monorepo.rst
```diff
@@ -0,0 +1,108 @@
+################################
```
raq929, line 1: Is there a reason we're using rst instead of `md` files here? (Like, don't change it, but I'm wondering why.)

docs/decisions/0002-lerna-monorepo.rst
```diff
@@ -0,0 +1,108 @@
...
+
+The image that follows is a diagram outlining the NPM publishing process. When a PR is merged, Github Actions will run the necessary Lerna commands to run all linting, tests, and builds for each package that changed since the last release, including dependencies and dependents. Each package, independently released, follows semantic versioning to determine the package version, creates and pushes Git tags, and then publishes to NPM.
+
+.. image:: https://user-images.githubusercontent.com/2828721/116780392-82947000-aa4a-11eb-82ce-819604de935f.png
+
+************
+Consequences
+************
+
+* The existing ``frontend-enterprise`` NPM package will be deprecrated in favor of the 3 new NPM packages outlined above. We will need to upgrade all uses of ``frontend-enterprise`` to import from the relevant individual NPM packages instead.
```
raq929, line 98: What's the plan to let people know that they have to upgrade and push them into actually doing it?

packages/catalog-search/README.rst
```diff
@@ -0,0 +1,5 @@
+@edx/frontend-enterprise-catalog-search
+==============
+
+This package exports UI components, React hooks, and actions functions related to Enterprise catalog search using `Algolia <algolia.com>`_ as a hosted search backend. This package will be used wherever it's
+necessary to allow users to search the catalog(s) available to them through their Enterprise Customer.
```
raq929, line 5: Link to the license.

packages/catalog-search/src/SearchFilters.jsx
```diff
@@ -1,22 +1,19 @@
 import React, { useMemo, useContext } from 'react';
 import PropTypes from 'prop-types';
 import { breakpoints } from '@edx/paragon';
+import { useWindowSize } from '@edx/frontend-enterprise-utils';
```
raq929, line 4: This hook is now in Paragon and is better tested and maintained there, should we take this opportunity to switch it out?

packages/catalog-search/src/tests/ClearCurrentRefinements.test.jsx
```diff
@@ -1,10 +1,10 @@
 import React from 'react';
+import { renderWithRouter } from '@edx/frontend-enterprise-utils';
```
raq929, line 2: Because `renderWithRouter` is only about rendering React components, I wonder if it should remain in this library? 
I could be convinced otherwise, thought.

packages/catalog-search/src/tests/SearchHeader.test.jsx
```diff
@@ -18,24 +19,24 @@ jest.mock('../SearchFilters', () => ({
 }));
 
 describe('SearchHeader', () => {
-  it('displays a SearchBox', () => {
-    render(<MemoryRouter><SearchData><SearchHeader /></SearchData></MemoryRouter>);
+  test('displays a SearchBox', () => {
```
raq929, line 22: Why are you changing `it` to `test` here?

raq929 (approved):
This looks great! I am super excited about it. I'm assuming you've tested this locally?
Some questions and nits, but no major concerns.

README.rst
adamstankiewicz: Yeah, I decided to remove these from the root README.rst in favor of adding them back into the individual package's README files once the new packages are published on NPM.

This will be a quick follow-up PR once things are published.

packages/catalog-search/src/SearchFilters.jsx
adamstankiewicz: I had a similar thought. If we are going to remove it, it probably makes sense to do now prior to first publish since it'll become a breaking change if we remove it once `@edx/frontend-enterprise-utils` is already published to NPM.

I will update!

packages/catalog-search/src/tests/ClearCurrentRefinements.test.jsx
adamstankiewicz: There will be 2 packages within the `frontend-enterprise` monorepo that export React components and both use `renderWithRouter`:

1. `@edx/frontend-enterprise-catalog-search`
2. `@edx/frontend-enterprise-logistration`

I opted to move `renderWithRouter` into `@edx/frontend-enterprise-utils` to be easily shared between libraries (or between our Enterprise MFEs).

docs/decisions/0002-lerna-monorepo.rst
adamstankiewicz: `.rst` seems to be the convention used elsewhere throughout edX repos for README files, so I'm just continuing with that convention here.

packages/catalog-search/src/tests/SearchHeader.test.jsx
adamstankiewicz: No reason in particular, to be honest. We seem to be using `test` more so than `it` lately so I swapped here for that reason. I can swap back to `it` or keep them as `test`. Do you have a strong opinion here?

renovate.json
```diff
@@ -2,6 +2,7 @@
   "extends": [
     "config:base"
   ],
+  "updateInternalDeps": true,
```
adamstankiewicz, line 5: According to the Renovate docs, this _should_ tell Renovate to update all the dependencies listed in `packages/**/package.json` files in addition to the `package.json` file in the repository root.

packages/catalog-search/src/tests/SearchHeader.test.jsx
raq929: Not really. I always use it because it encourages me to write the test name like a sentence, but I don't actually care much.

.github/workflows/release.yml
```diff
@@ -1,29 +1,43 @@
...
-      - name: Release
+        run: npm run ci:test
+      - name: Coverage Report
+        uses: codecov/codecov-action@v1
+      - name: Build
+        run: npm run ci:build
+      - name: Preview Changed Packages
+        run: npm run changed
+      - name: Publish Changed Packages
+        run: npm run publish
```
iloveagent57, line 41: Nit: further down, in the ADR, you talk about `learna publish` - is `npm run publish` aliasing to `learna publish` since learna is installed/configured?

iloveagent57 (approved):
I am in favor of this change and have one question about the `publish` script.

docs/decisions/0002-lerna-monorepo.rst
```diff
@@ -0,0 +1,108 @@
...
+necessary to allow users to search the catalog(s) available to them through their Enterprise Customer.
+
+--------------------------------
+``@edx/frontend-enterprise-logistration``
+--------------------------------
+
+This package exports a UI component related to logistration for Enterprise users. This package will be helpful for Enterprise micro-frontends to share logic related to redirecting enterprise users to the logistration page branded with their Enterprise Customer's logo and additional messaging.
+
+--------------------------------
+``@edx/frontend-enterprise-utils``
```
taliaedX, line 59: what is the likelihood this becomes a catch-all package?

Ie I have a new shared component/function/thing and maybe it's not explicitly or obviously related to catalog search or logistration? I just imagine this package could eventually run into the same issues the overall repo did (too big, too many irrelevant dependencies, etc.) and why we split it into 3 packages. Do we plan to have any guidelines on that? Or is that a decision we make at some point if we see it getting too complicated?

commitlint.config.js
```diff
@@ -1,3 +1,3 @@
 module.exports = {
-  extends: ['@commitlint/config-angular'],
+  extends: ['@commitlint/config-conventional'],
```
binodpant, line 2: 🎉

docs/decisions/0002-lerna-monorepo.rst
```diff
@@ -0,0 +1,108 @@
...
+--------------------------------
+
+This package exports useful React hooks, utility functions, and constants for use in other libraries and micro-frontends, including checking whether an authenticated user is an enterprise user based on their JWT role(s), or fetching a user's available Enterprise learner portal links.
+
+================================
```
binodpant, line 64: post this merged, I would love for this content to be part of the main README as well, it's pretty important stuff to avoid dependency confusions

docs/decisions/0002-lerna-monorepo.rst
```diff
@@ -0,0 +1,108 @@
...
+NPM releases with Lerna
+================================
+
+Publishing packages to NPM remains a job of Github Actions, however, releases will no longer happen using ``semantic-release`` directly. Instead, we will rely on the ``lerna publish`` command to determine which packages have changed since the last merged commit and release, and only publish those. 
+
+Lerna is configured to use independent versioning with conventional commits, as opposed to keeping all package versions in sync. This will allow us to publish packages independently of each other, rather than in lock-step like some JS monorepos do (e.g., Babel, React). The benefit of independent versioning is that consuming applications won't get erroneous upgrades when code unrelated to their use case is updated in the monorepo. For example, say the ``edx/frontend-app-learning`` micro-frontend consumes ``@edx/frontend-enterprise-utils``. Without independent versioning, if we released a new version of ``@edx/frontend-enterprise-catalog-search``, then ``edx/frontend-app-learning`` would have an upgrade available when nothing about the package changed.
```
binodpant, line 88: If a change to utils module is made and it only breaks logistration but not catalog-search, for example, will the deployment of utils module also be gated, or would it deploy regardless?

docs/decisions/0002-lerna-monorepo.rst
```diff
@@ -0,0 +1,108 @@
...
+
```
binodpant, line 89: If a change to utils module is made and it only breaks logistration but not catalog-search, for example, will the deployment of utils module also be gated, or would it deploy regardless?

packages/catalog-search/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "@testing-library/react-hooks": "3.4.2",
+    "@testing-library/user-event": "13.1.5",
+    "react-instantsearch-dom": "6.8.3"
+  },
+  "peerDependencies": {
+    "@edx/frontend-platform": "1.9.4",
+    "@edx/paragon": "14.5.0",
+    "@fortawesome/free-solid-svg-icons": "5.8.1",
+    "@fortawesome/react-fontawesome": "0.1.4",
+    "react": "16.14.0",
```
binodpant, line 53: Thinking of a case where hypothetically two of the modules here use two different versions of react, for example. Would we handle that somehow?

This used to be possible in Maven, but it would be nice there was some way to constraint/single-source the version of some of these dependencies like react version, at the topmost level so that all the packages in this repository would ideally use the same version. 

The concept of "managed deps" allowed to do that in maven/gradle (I think gradle calls its aligning versions https://docs.gradle.org/current/userguide/dependency_version_alignment.html)

In the case of maven, the main package file equivalent would have the version, and the dependent module only asked for a dependency but without a version, since the version was managed by the main package file. This meant that the modules never went out of sync wrt important deps such as react in this case, ensuring compatibility. 

It probably won't be a big issue since we can just be vigialnt to update each module if there are major updates to say react, and I don't even know if npm allows for that. But we might be able to get the equivalent effect somehow.

packages/utils/package.json
```diff
@@ -0,0 +1,47 @@
...
+  "homepage": "https://github.com/edx/frontend-enterprise#readme",
+  "publishConfig": {
+    "access": "public"
+  },
+  "bugs": {
+    "url": "https://github.com/edx/frontend-enterprise/issues"
+  },
+  "sideEffects": false,
+  "dependencies": {
+    "@testing-library/react": "11.2.6",
```
binodpant, line 34: should this be dep or devDep?

docs/decisions/0002-lerna-monorepo.rst
```diff
@@ -0,0 +1,108 @@
...
+.. image:: https://user-images.githubusercontent.com/2828721/116780392-82947000-aa4a-11eb-82ce-819604de935f.png
+
+************
+Consequences
+************
+
+* The existing ``frontend-enterprise`` NPM package will be deprecrated in favor of the 3 new NPM packages outlined above. We will need to upgrade all uses of ``frontend-enterprise`` to import from the relevant individual NPM packages instead.
+* Common or shared dependencies will be accessible to all packages in the monorepo, regardless of whether they are explicitly specified in a package's package.json file. Despite this, we have tooling in place through ESlint's ``eslint-plugin-import`` to throw a linting error when attempting to import packages that are not explicitly defined in package.json.
```
binodpant, line 99: does this mean that if I introduce a dep into the top level, and a specific package does not need/use it, what will happen? Will it still be in the dependency tree of that package? I was hoping there would be a way to say, only make a dep visible to one of the packages, if they explicitly depend on it, otherwise no.

binodpant (approved):
no flags, but a few discussion questions on deps and workflows!

docs/decisions/0002-lerna-monorepo.rst
adamstankiewicz: Agreed! Once packages get published, I'll be doing a quick follow-up PR to update README files to include links to things that don't exist yet until they are published.

While I'm doing that, I'll be sure to bring some of this content into the main README.

packages/utils/package.json
adamstankiewicz: It is currently a regular dependency and after some consideration, I left it as a regular dependency due to that the `renderWithRouter` test utility function we have in the `frontend-enterprise-utils` package relies on `@testing-library/react` as an implementation detail, so to ensure it gets bundled when publishing to NPM, it will need to remain a dependency.

docs/decisions/0002-lerna-monorepo.rst
adamstankiewicz: Say we have some new package that doesn't rely on React. That package would not define anything to do with React in its own package.json file, and as such, the NPM distribution won't have any reference or reliance on React; similarly, React wouldn't appear in that package's `node_modules`.

However, because React _is_ installed as a devDep in the root package.json file, it will still be accessible to any package within the monorepo even if it's not explicitly declared in a package's package.json file. This is due to how NPM resolves `node_modules` when importing; it continues to search up the file system tree to find "node_modules" folders in parent directories.

While we can't prevent that import from happening, we will at least throw a linting error (which fails CI) when trying to use a package from the root `node_modules` that isn't explicitly defined in an individual package's package.json file.

See https://github.com/lerna/lerna/blob/main/doc/hoist.md#forgetting-dependencies for more details. This document is specifically about `lerna bootstrap --hoist` which literally takes _all_ shared dependencies across all packages in the monorepo and installs them in the root, but `--hoist` felt too much like magic, especially since we may not want to always "hoist" up all shared dependencies. The only time I think it makes sense to hoist a package is when it will run into conflicts by having 2 different versions in use at the same time (like react, react-dom, etc.). But the same principle applies; we're just sort of being explicit about what to hoist and what not to hoist.

packages/catalog-search/package.json
adamstankiewicz: From a development point of view, there is a constraint / single source of truth for version of React since it is hoisted up into the root `node_modules`. Any package that relies on React, locally during development, would import from this single React version.

It gets a little trickier with the peer dependencies. I think we should switch to using semantic version ranges in our peer dependencies (follow-up ticket TBD) to say that "Hey, this package supports React 16.0.6 all the way through React 17", for example, which would help here. Since the consuming application (e.g., learner portal MFE) is providing the actual React dependency due to the fact it's a peer dependency, so long as our packages keep React as a peer dependency with valid ranges of compatible versions, we should be OK here.

docs/decisions/0002-lerna-monorepo.rst
adamstankiewicz: Good call out, @taliaedX! We definitely run the risk of it becoming a catch-all package and start running into similar issues. I agree we should add more explicit guidance around this somewhere, perhaps a section in the README in the root of the monorepo that provides some rough guidelines / considerations to make on whether to add new things to existing packages or start a new package (and how to do so).

I might bucket that into the next follow-up PR when updating some of the READMEs anyways?

docs/decisions/0002-lerna-monorepo.rst
adamstankiewicz: The Github Actions CI runs tests for each package that was changed since the last release and any of its dependencies and dependents. So, if we change `frontend-enterprise-utils` and break something, CI _should_ fail if it causes `frontend-enterprise-catalog-search` and/or `frontend-enterprise-logistration` tests to fail since both of those packages are dependents on `frontend-enterprise-utils`.

But yeah, this is actually something I still need to explicitly test to see it holds up to what should conceptually happen. I will try to replicate this situation locally and report back in the morning!

.github/workflows/release.yml
adamstankiewicz: Yeah, that's correct. There's a "publish" NPM script defined in the root package.json file that `npm run publish` calls, which runs `lerna publish`.

docs/decisions/0002-lerna-monorepo.rst
adamstankiewicz: Confirmed. I caused a bug in `@edx/frontend-enterprise-utils` in a function that's imported by `@edx/frontend-enterprise-catalog-search`, and the `npm run ci:test` ended up running both `@edx/frontend-enterprise-catalog-search` and `@edx/frontend-enterprise-logistration` (since both depend on `@edx/frontend-enterprise-utils`), and the test suite for `@edx/frontend-enterprise-catalog-search` failed due to the regression made, which would fail CI and prevent publishing of the new version of `@edx/frontend-enterprise-utils`.

PR #100, opened 2021-05-06
fix: updates to release config for Github Actions

adamstankiewicz:
On the first attempt to publish to the 3 packages to NPM, the running of lint, test, and build commands was previously only running packages that have changed since `origin/master` to avoid running things for packages that did not change.

However, since the refactor PR merged to `master`, the commit is equal with `origin/master` so Lerna deems that no changes have occurred:

image

Also, it turns out `lerna publish` does not accept the `--stream`` argument so this PR also removes that.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,13 +28,13 @@ jobs:
       - name: Run Lerna Bootstrap
         run: npm run bootstrap
       - name: Lint
-        run: npm run ci:lint
+        run: npm run lint
       - name: Test
-        run: npm run ci:test
+        run: npm run test
       - name: Coverage Report
         uses: codecov/codecov-action@v1
       - name: Build
-        run: npm run ci:build
+        run: npm run build
       - name: Preview Changed Packages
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -28,13 +28,13 @@ jobs:
       - name: Run Lerna Bootstrap
         run: npm run bootstrap
       - name: Lint
-        run: npm run ci:lint
+        run: npm run lint
```
adamstankiewicz, line 31: Running `npm run lint` vs. `npm run ci:lint` will run linting for _all_ packages vs. just the ones that were changed since `origin/master`. However, when a PR is merged to `master`, it's commit is even with `origin/master` so the commands weren't being run.

I'm opting to run these commands on merge to `master` for all packages in the monorepo, which is fine for now since there's only 3 but we may want to come back and update if releases start taking longer as this monorepo grows.

Similar for the other below commands, too.

iloveagent57 (approved):

PR #104, opened 2021-05-06
docs: update readme files

adamstankiewicz:
Now that our 3 packages are successfully published to NPM, I've updated their individual README files to include badges (e.g., NPM version and license info) related to the published packages.

I also brought some content from the ADR into the main README file and updates some its content.


packages/catalog-search/package.json
```diff
@@ -10,7 +10,17 @@
...
+  "watch": {
+    "build": {
+      "patterns": [
+        "src"
+      ],
+      "extensions": "js,jsx"
+    }
+  },
   "scripts": {
+    "dev": "npm-watch build",
```
adamstankiewicz, line 22: This NPM script begins watching for any changes in the `src` files matching extensions `.js` and `.jsx` and executes `npm run build` on any of these changes. This is used alongside with Lerna to watch all package source files and re-transpile as needed to ensure `dist` directories always exist in symlinking.

packages/catalog-search/package.json
```diff
@@ -23,7 +33,8 @@
   "license": "AGPL-3.0",
   "homepage": "https://github.com/edx/frontend-enterprise#readme",
   "publishConfig": {
-    "access": "public"
+    "access": "public",
+    "directory": "dist"
```
adamstankiewicz, line 37: This config option tells Lerna to only publish the contents of the `dist` folder, rather than all the files. The `make build` command ensures all the necessary files are added to the `dist` folder at build time.

.github/workflows/release.yml
```diff
@@ -24,10 +24,11 @@ jobs:
...
           npm whoami
-      - name: Install Dependencies
-        run: npm ci
-      - name: Run Lerna Bootstrap
-        run: npm run bootstrap
+      - name: Install and Setup Dependencies
+        run: npm run ci:setup
+      # build must come before running linting and tests for the `dist` directory to exist.
+      - name: Build
+        run: npm run ci:build
```
iloveagent57, line 31: There's still another "Build" job below.

packages/catalog-search/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@edx/frontend-enterprise-catalog-search",
-  "version": "0.1.4",
+  "version": "0.1.0",
```
iloveagent57, line 3: Did you want to go back to 0.1.0?

adamstankiewicz: Ooooh, so I had mistakenly thought these were bumped from running `npm pack` without the `--dry-run` option. But, actually there were actually automated releases due to package upgrades by Renovate. I've disabled Renovate's automerge on patch releases for the time being.

Great catch!

iloveagent57 (approved):

PR #113, opened 2021-05-11
ci: when bootstraping packages in CI, also include dependencies and dependents

adamstankiewicz:


.github/workflows/release.yml
```diff
@@ -35,8 +35,6 @@ jobs:
         run: npm run test
       - name: Coverage Report
         uses: codecov/codecov-action@v1
-      - name: Preview Changed Packages
-        run: npm run changed
```
adamstankiewicz, line 39: I previously removed this from ci.yml, but realized we should also remove it from release.yml since `lerna publish` is already going to only publish changed packages, there's no needed to fail CI if no packages have changed (as is the case in PRs like this one).

package.json
```diff
@@ -11,7 +11,7 @@
     "setup": "npm install && npm run bootstrap",
     "ci:setup": "npm install && npm run ci:bootstrap",
     "bootstrap": "lerna bootstrap",
-    "ci:bootstrap": "lerna bootstrap --since origin/master",
+    "ci:bootstrap": "lerna bootstrap --include-dependencies --include-dependents --since origin/master",
```
adamstankiewicz, line 14: Adding the `--include-dependencies` and `--include-dependents` flags here were mistakenly left out and didn't match the other CI commands like:

```
"ci:build": "lerna run build --include-dependencies --include-dependents --since origin/master --parallel",
```

PR #114, opened 2021-05-11
ci: use from-package arg in lerna publish

adamstankiewicz:
After another frontend-enterprise PR was merged in earlier today, it was observed that Lerna attempted to create a new release for `@edx/frontend-enterprise-catalog-search@0.1.7` by pushing a "chore: publish" commit to update the package.json version as well as pushing the `0.1.7` tag.

However, `lerna publish` failed to publish 0.1.7 to the NPM registry as the publish command seemed to attempt to publish 0.1.6 instead, which correctly errors since that version is already published on the NPM registry. As far I can tell, this is happening because the `lerna publish` isn't recognizing the changes to package.json versions (from the generated "chore: publish" commit) when publishing to NPM.

As a result, I separated the versioning from the publishing as two distinct steps:
1. First, `lerna version` is run to update the package.json version(s), add CHANGELOG entr(ies), create new Git tag(s). All these changes are pushed to the Git remote.
2. Second, `lerna publish` is now run with the `from-package` positional argument to tell Lerna to publish from the versions explicitly specified in the package.json files from the "chore: publish" commit from `lerna version`.


.github/workflows/ci.yml
```diff
@@ -23,13 +23,21 @@ jobs:
...
-        run: npm run ci:test
+        run: lerna run test --include-dependencies --include-dependents --since origin/master --parallel
       - name: Coverage Report
         uses: codecov/codecov-action@v1
+      - name: Extract branch name
+        shell: bash
+        run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/})"
+        id: extract_branch
+      - name: Preview Updated Versions (dry run)
+        run: lerna version --allow-branch ${{ steps.extract_branch.outputs.branch }} --conventional-commits --no-git-tag-version --no-changelog --no-push --yes
```
adamstankiewicz, line 43: Lerna is configured to only allow versioning/publishing on the `master` branch. However, by passing in the branch name explicitly to `--allow-branch`, we can override this to preview a dry-run of the version command.

.github/workflows/release.yml
```diff
@@ -35,7 +35,9 @@ jobs:
...
       - name: Coverage Report
         uses: codecov/codecov-action@v1
-      - name: Publish Changed Packages
-        run: npm run publish
+      - name: Push Updated Versions and Tags
+        run: lerna version --yes
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Publish Changed Packages
+        run: lerna publish from-package --yes
```
adamstankiewicz, line 43: `from-package` here should tell Lerna to get package versions from the package.json changes that were made in the previous step for `lerna version`, which are already committed to the rep at this point.

iloveagent57 (approved):

PR #119, opened 2021-05-14
fix: add hotjar to searchbox

raq929:
Adds the necessary tag for hotjar analytics.
Verified manually that the tag appears by using local `frontend-enterprise` on learner portal.

### DIFF
--- a/packages/catalog-search/src/SearchBox.jsx
+++ b/packages/catalog-search/src/SearchBox.jsx
@@ -66,7 +66,7 @@ export const SearchBoxBase = ({
         onSubmit={handleSubmit}
         onClear={handleClear}
       >
-        <SearchField.Input className="form-control-lg" aria-labelledby="search-input-box" />
+        <SearchField.Input className="form-control-lg" aria-labelledby="search-input-box" data-hj-whitelist />
         <SearchField.ClearButton />
         <SearchField.SubmitButton />
       </SearchField.Advanced>


packages/catalog-search/src/SearchBox.jsx
```diff
@@ -66,7 +66,7 @@ export const SearchBoxBase = ({
         onSubmit={handleSubmit}
         onClear={handleClear}
       >
-        <SearchField.Input className="form-control-lg" aria-labelledby="search-input-box" />
+        <SearchField.Input className="form-control-lg" aria-labelledby="search-input-box" data-hj-whitelist />
```
taliaedX, line 69: does this function the same way as just appending it to an HTML class though? I'm unclear if the searchField component picks up extraneous variables and treat them as though they are HTML attributes.

taliaedX:
```
Using element attributes:


<input type=”text” name=”company” data-hj-whitelist />

<textarea name=”comments” data-hj-whitelist></textarea>

Using element classes:


<input type=”text” name=”company” class=”data-hj-whitelist” />

<textarea name=”comments” class=”data-hj-whitelist”></textarea>
```
How does a component level attribute get translated? 🤔

raq929: In the DOM it appears as
```
<input class="form-control form-control-lg" type="text" aria-labelledby="search-input-box" data-hj-whitelist="true" role="searchbox" id="pgn-searchfield-input-1" name="searchfield-input" value="">
```

taliaedX: interesting. So it gets read as "attribute that exists and thus is true". Weird. Okay.

taliaedX (approved):
Now understanding how the attribute gets passed through comnpletely from the component to the HTML element means this looks good to me!

packages/catalog-search/src/SearchBox.jsx
taliaedX: I expected components to only accept HTML attributes that were like builtins I guess? like class and such or to expect this tag to be part of the class. But then I am not sure how aria attributes would so cleanly get through (unless they are also in the actual base HTML spec)

PR #122, opened 2021-05-17
fix: split role name from context in JWT user roles

adamstankiewicz:


packages/utils/src/roles.js
```diff
@@ -28,12 +28,14 @@ export const isEnterpriseRole = role => ENTERPRISE_PERMISSIONS.some(permission =
  * @returns {boolean} true if the user has an enterprise role and/or the specified role, otherwise false
  */
 export const isEnterpriseUser = (user, role) => {
```
raq929, line 30: I'd love it if there were tests for this.

raq929 (approved):

packages/utils/src/roles.js
adamstankiewicz: Tests have been added for the functions in roles.js 👍

iloveagent57 (approved):

PR #132, opened 2021-06-09
maint: adding New Relic data tags for monitoring purposes

alex-sheehan-edx:
[Jira](https://openedx.atlassian.net/browse/ENT-4500)

We want to allow for slightly less fragile scripted monitoring systems by adding New Relic specific data attributes that will indicate to devs that changing or removing may cause alerts.

### DIFF
--- a/packages/catalog-search/src/SearchBox.jsx
+++ b/packages/catalog-search/src/SearchBox.jsx
@@ -52,6 +52,9 @@ export const SearchBoxBase = ({
     dispatch(deleteRefinementAction(QUERY_PARAM_FOR_PAGE));
   };
 
+  // Changing or removing New Relic synthetic data attributes will trigger a failure alert on our synthetics tests
+  // these should be disabled before removing the data attrs.
+  // https://docs.newrelic.com/docs/synthetics/synthetic-monitoring/scripting-monitors/introduction-scripted-browser-monitors/
   return (
     <div className={className}>
       {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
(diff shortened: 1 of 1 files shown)


packages/catalog-search/src/SearchBox.jsx
```diff
@@ -66,9 +69,9 @@ export const SearchBoxBase = ({
         onSubmit={handleSubmit}
         onClear={handleClear}
       >
-        <SearchField.Input className="form-control-lg" aria-labelledby="search-input-box" data-hj-whitelist />
-        <SearchField.ClearButton />
-        <SearchField.SubmitButton />
+        <SearchField.Input className="form-control-lg" aria-labelledby="search-input-box" data-hj-whitelist data-nr-synth-id="input-field" />
+        <SearchField.ClearButton data-nr-synth-id="clear-button" />
+        <SearchField.SubmitButton data-nr-synth-id="submit-button" />
```
adamstankiewicz, line 74: Curious if the attribute values here should be a bit more specific to Catalog Search? I'm thinking there could potentially be some other submit button or input field on the page outside of the catalog-search's `SearchBox` component and if it also had generic "submit-button" or "input-field", would NewRelic be able to distinguish between these?

alex-sheehan-edx: Hmm, good point. I believe there's a chance for that situation, and while theoretically would be possible to distinguish within New Relic, it would add a bunch more complexity needed within the Synthetics test which could easily be solved by simply adding a `catalog-search` to the beginning of the ID's here- so I'll add

raq929 (approved):
Lgtm!

PR #145, opened 2021-08-13
fix: improved support for lists of attribute values in Algolia facet filtering

adamstankiewicz:
This PR refactors how the Algolia search refinements are synced with the URL query parameters, and fixes a bug where a refinement with a command in it would break the search results.

On initial page load, it parses the URL query parameters and applies them to the Algolia search refinements. However, previously, when a new refinement is selected, we would add it to the refinements reducer state and `SearchContext` would respond to this change, and attempt to update the query parameters.

This causes issues because the logic that parses the initial URL query parameters was also attempting to parse updates made to the query params when the refinements reducer changed. As a result, we were in a situation where these two bits of logic were conflicting and causing issues.

This PR updates this logic to only parse the query parameters to update the refinements reducer state on initial component mount, instead of any time the query parameter string changed. This removes the conflicting logic and maintains a "source-of-truth" for which search state should be synced with URL query parameters.

**Note: This is a breaking change!**

### Testing instructions

1. Checkout this branch of `frontend-enterprise` locally.
2. Checkout latest `frontend-app-learner-portal-enterprise`.
3. Add `@edx/frontend-enterprise-catalog-search` to the Learner Portal module.config.js file:
```js
{
  moduleName: '@edx/frontend-enterprise-catalog-search',
  dir: '../frontend-enterprise/packages/catalog-search',
  dist: 'src',
},
```
4. Add the production Algolia API keys in .env.private in Learner Portal and update `useDefaultSearchFilters` to return the following filter string: `enterprise_customer_uuids:8ea70d02-c210-4009-a86d-e23423dcda7d`. This will ensure the search results return with refinements that have commas (e.g., "University of Maryland, College Park").
5. Run `npm start` in Learner Portal. and browse to the search page (e.g., https://localhost:8734/test-enterprise/search).
6. Select the "University of Maryland, College Park" option under the "Partner" facet and observe the full refinement label is shown as a badge rather than being split on the comma.
7. Select another partner while leaving the previous one selected; observe the search results update accordingly and an additional query parameter is added to the URL as appropriate.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -10,5 +10,9 @@
       "conventionalCommits": true,
       "createRelease": "github"
     }
+  },
+  "changelogPreset": {
+    "name": "conventionalcommits",
+    "issueUrlFormat": "{{host}}/{{owner}}/{{repository}}/issues/{{id}}"
   }
 }
--- a/packages/catalog-search/__mocks__/react-instantsearch-dom.jsx
+++ b/packages/catalog-search/__mocks__/react-instantsearch-dom.jsx
@@ -49,7 +49,7 @@ MockReactInstantSearch.connectRefinementList = Component => (props) => (
     attribute="subjects"
     currentRefinement={[]}
     items={[]}
-    refinementsFromQueryParams={{}}
+    refinements={{}}
     title="Foo"
     searchForItems={() => {}}
     {...props}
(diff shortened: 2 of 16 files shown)


iloveagent57 (approved):
LGTM

PR #146, opened 2021-08-16
feat: ENT4816- Add jobs dropdown

zamanafzal:
The objective of this PR is to add more options to current Algolia widgets.
Newly added functionality in this PR:
1-Dropdown state in context was saved against attribute which can now be changed and saved against `customAttrbiute,` so multiple dropdowns can be loaded against the same attribute.
2- If it is desired to not refine the data from Algolia on the basis of user selection but only mark the checkboxes and save his selections.
3- Previously `SEARCH_FACET_FILTERS` selections were only saved as an array. Any other attributes value is saved as a string. Now if it is needed to save any other attribute value in the array then its key can be added in `ADDITIONAL_FACET_FILTERS.

https://openedx.atlassian.net/browse/ENT-4843
https://openedx.atlassian.net/browse/ENT-4844

Extra Details:
**doRefinement**:
We currently have a use case where we do not want to refine the data on the basis of user selections from `facet` but to save the user selection and show these checkboxes marked in the dropdown.
For that, I have added `doRefinement` flag which will be True by default and will work as it was working previously. But if this flag is set to false then checkboxes state will not be marked from algolia's `isRefined` flag and checked from the local state.
**customAttribute**:
We want to load two dropdowns for jobs from the same attribute and keep their state different. Since the state is saved against the attribute name, so to change the attribute name an optional `customAttribute` can be passed.

### DIFF
--- a/packages/catalog-search/src/FacetListBase.jsx
+++ b/packages/catalog-search/src/FacetListBase.jsx
@@ -21,6 +21,8 @@ const FacetListBase = ({
   searchForItems,
   variant,
   noDisplay,
+  doRefinement,
+  customAttribute,
 }) => {
   const { refinements, dispatch } = useContext(SearchContext);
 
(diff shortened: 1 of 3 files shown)


packages/catalog-search/src/FacetListBase.jsx
```diff
@@ -54,9 +59,14 @@ const FacetListBase = ({
       if (!items?.length) {
         return <span className="p-2 d-block">{NO_OPTIONS_FOUND}</span>;
       }
-
+      let isChecked = false;
       return items.map(item => {
-        const isChecked = isCheckedField ? item[isCheckedField] : !!item.value;
+        if (doRefinement) {
```
zamanafzal, line 64: `doRefinement` will always be `True` as per L119 and there will be no change in the existing flow except for the dropdowns where we want to not refine the data from Algolia.

packages/catalog-search/src/FacetListBase.jsx
```diff
@@ -31,19 +33,22 @@ const FacetListBase = ({
   const { refinementsFromQueryParams, dispatch } = useContext(SearchContext);
 
   const handleInputOnChange = (item) => {
-    if (item.value && facetValueType === 'array') {
+    // if it is desired to load the same attribute data in multiple dropdowns then
+    // customAttribute can be passed to differentiate them.
+    const index = customAttribute || attribute;
```
zamanafzal, line 38: The default value of `customAttribute` is null as per L115. So there will be no change in existing behaviour.

packages/catalog-search/src/data/constants.js
```diff
@@ -51,6 +51,8 @@ export const SEARCH_FACET_FILTERS = [
   },
 ];
 
+export const ADDITIONAL_FACET_FILTERS = ['skills.name', 'job_name'];
```
sameenfatima78, line 54: Facet filter for job is now `name`

zamanafzal: Yes, you are right. But I've requested @saleem-latif to change it to job_name, as in URL it doesn't make sense what name is it.

saleem-latif (approved):

sameenfatima78 (approved):
Good to go! 👍

muhammad-ammar (approved):
LGTM

PR #147, opened 2021-08-16
fix: Updated FacetListBase to save list of selected item

HammadAhmadWaqas:
Updated FacetListBase to save a list of the selected items even for `single-item` face-value-type.
At many places we assume that we will get a list of items, otherwise fails. to keep that consistent with the `array` type, setting the array with one value.


muhammad-ammar (approved):
LGTM

PR #157, opened 2021-08-17
ci: pull latest tags after versioning

adamstankiewicz:


binodpant (approved):
yeah let's try this or `git fetch --all --tags` ?

PR #215, opened 2022-02-17
feat: add type ahead autocomplete suggestion support

muneebGH:
original approach mentioned in discovery was to use [this](https://www.algolia.com/doc/api-reference/widgets/autocomplete/react/) approach but we can see we have to call `refine` on every onChange which resulted in updating the results of whole page (instead of suggestions only) which was not desire-able.

So i went on with calling algolia directly through client being used in learner portal.

https://user-images.githubusercontent.com/49611774/154471057-6f322923-7a35-4652-bbae-552640aadb04.mov

**Merge checklist:**
- [x] Ensure your commit message follows the semantic-release conventional commit message format
**Post merge:**


packages/catalog-search/src/SearchSuggestions.jsx
```diff
@@ -0,0 +1,62 @@
...
+      {
+                autoCompleteHits.filter(hit => hit.content_type === 'course')
+                  .slice(0, 3)
+                  .map((hit) => (
+                    <Link to={`/${enterpriseSlug}/course/${hit.key}`} key={hit.title} className="suggestion-item" style={{ whiteSpace: 'pre-wrap' }}>
+                      <div style={{ display: 'flex' }}>
+                        {/* eslint-disable-next-line no-underscore-dangle, react/no-danger */}
+                        <div dangerouslySetInnerHTML={{ __html: hit._highlightResult.title.value }} />
+                        <div className="badge badge-light ml-3 font-weight-light " style={{ lineHeight: '1.5' }}>
+                          {hit.key && hit.key.split('+')[0]}
```
sameenfatima78, line 20: Aren't we supposed to show the complete course/program key here?

packages/catalog-search/src/SearchSuggestions.jsx
```diff
@@ -0,0 +1,62 @@
...
+
+const SearchSuggestions = ({ autoCompleteHits, enterpriseSlug, handleViewAllClick }) => (
+  <div className="suggestions" data-testid="suggestions">
+    <div>
+      <div className="mb-2 ml-2 mt-1 font-weight-bold suggestions-section">
+        Courses
+      </div>
+      {
+                autoCompleteHits.filter(hit => hit.content_type === 'course')
+                  .slice(0, 3)
```
sameenfatima78, line 13: nit: use a constant here instead of hardcoded value, 
```javascript
slice(0, MAX_NUM_SUGGESTIONS)

File: packages/catalog-search/src/styles/_SearchBox.scss [hunk 3]
```diff
@@ -5,3 +5,55 @@
...
+  margin-top: auto;
+  z-index: 1;
+  box-shadow: 0 4px 8px 0 rgba(0, 0, 0, 0.2), 0 6px 20px 0 rgba(0, 0, 0, 0.19);
+
+}
+
+.suggestion-item {
+  display: flex;
+  flex-direction: column;
+ margin: 5px 10px 5px 10px;
```
sameenfatima78, line 22: nit: add one space indentation here

packages/catalog-search/src/SearchSuggestions.jsx
```diff
@@ -0,0 +1,62 @@
...
+      </div>
+      {
+                autoCompleteHits.filter(hit => hit.content_type !== 'course')
+                  .slice(0, 3)
+                  .map((hit) => (
+                    <Link to={`/${enterpriseSlug}/program/${hit.aggregation_key.split(':').pop()}`} key={hit.title} className="suggestion-item" style={{ whiteSpace: 'pre-wrap' }}>
+                      <div style={{ display: 'flex' }}>
+                        {/* eslint-disable-next-line no-underscore-dangle, react/no-danger */}
+                        <div dangerouslySetInnerHTML={{ __html: hit._highlightResult.title.value }} />
+                        <div className="badge badge-light ml-3 font-weight-light " style={{ lineHeight: '1.5' }}>
```
sameenfatima78, line 39: nit: remove extra space from `font-weight-light "`

packages/catalog-search/src/SearchSuggestions.jsx
```diff
@@ -0,0 +1,62 @@
...
+    </div>
+    <div>
+      <div className="mb-2 mt-5 ml-2 font-weight-bold suggestions-section">
+        Programs
+      </div>
+      {
+                autoCompleteHits.filter(hit => hit.content_type !== 'course')
+                  .slice(0, 3)
+                  .map((hit) => (
+                    <Link to={`/${enterpriseSlug}/program/${hit.aggregation_key.split(':').pop()}`} key={hit.title} className="suggestion-item" style={{ whiteSpace: 'pre-wrap' }}>
```
sameenfatima78, line 35: Same suggestion as above: 

```javascript
<Link
   key={hit.title}
   className="suggestion-item"
   style={{ whiteSpace: 'pre-wrap' }}
   to={linkToProgram}
 >

```

packages/catalog-search/src/SearchSuggestions.jsx
```diff
@@ -0,0 +1,62 @@
...
+  <div className="suggestions" data-testid="suggestions">
+    <div>
+      <div className="mb-2 ml-2 mt-1 font-weight-bold suggestions-section">
+        Courses
+      </div>
+      {
+                autoCompleteHits.filter(hit => hit.content_type === 'course')
+                  .slice(0, 3)
+                  .map((hit) => (
+                    <Link to={`/${enterpriseSlug}/course/${hit.key}`} key={hit.title} className="suggestion-item" style={{ whiteSpace: 'pre-wrap' }}>
```
sameenfatima78, line 15: I think the code will a lot more elegant if we pass a reference to a function here that will return us the path:

```javascript
<Link
   key={hit.title}
   className="suggestion-item"
   style={{ whiteSpace: 'pre-wrap' }}
   to={linkToCourse}
 >

```

sameenfatima78 (commented):
Nice work... just a few suggestions. Also, would be nice if we can get @adamstankiewicz's review on your PR.

packages/catalog-search/src/SearchSuggestions.jsx
muneebGH: this is basically to fetch name of authoring organization embedded in key, such as "HarvardX+CS50W", we would need only HarvardX from here. so i split this and choose relevant part.

packages/catalog-search/src/SearchSuggestions.jsx
muneebGH: done

packages/catalog-search/src/SearchSuggestions.jsx
```diff
@@ -0,0 +1,69 @@
...
+          Courses
+        </div>
+        {
+                    autoCompleteHits.filter(hit => hit.content_type === 'course')
+                      .slice(0, MAX_NUM_SUGGESTIONS)
+                      .map((hit) => (
+                        <Link to={getLinkToCourse(hit)} key={hit.title} className="suggestion-item" style={{ whiteSpace: 'pre-wrap' }}>
+                          <div style={{ display: 'flex' }}>
+                            {/* eslint-disable-next-line no-underscore-dangle, react/no-danger */}
+                            <div dangerouslySetInnerHTML={{ __html: hit._highlightResult.title.value }} />
```
zamanafzal, line 24: Is it ok to do it this way? As it can open up a cross-site scripting (XSS) attack.

packages/catalog-search/src/SearchSuggestions.jsx
```diff
@@ -0,0 +1,69 @@
+import { Link } from 'react-router-dom';
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { MAX_NUM_SUGGESTIONS } from './data/constants';
+
+const SearchSuggestions = ({ autoCompleteHits, enterpriseSlug, handleViewAllClick }) => {
+  const getLinkToCourse = (course) => `/${enterpriseSlug}/course/${course.key}`;
+  const getLinkToProgram = (program) => `/${enterpriseSlug}/program/${program.aggregation_key.split(':').pop()}`;
```
zamanafzal, line 9: why we are using pop function here?

packages/catalog-search/src/SearchSuggestions.jsx
```diff
@@ -0,0 +1,69 @@
...
+        </div>
+        {
+                    autoCompleteHits.filter(hit => hit.content_type === 'course')
+                      .slice(0, MAX_NUM_SUGGESTIONS)
+                      .map((hit) => (
+                        <Link to={getLinkToCourse(hit)} key={hit.title} className="suggestion-item" style={{ whiteSpace: 'pre-wrap' }}>
+                          <div style={{ display: 'flex' }}>
+                            {/* eslint-disable-next-line no-underscore-dangle, react/no-danger */}
+                            <div dangerouslySetInnerHTML={{ __html: hit._highlightResult.title.value }} />
+                            <div className="badge badge-light ml-3 font-weight-light" style={{ lineHeight: '1.5' }}>
```
zamanafzal, line 25: Except of adding styles this way, It is recommended to add class in stylesheet.

packages/catalog-search/src/SearchSuggestions.jsx
```diff
@@ -0,0 +1,69 @@
...
+                              {hit.key && hit.key.split('+')[0]}
```
zamanafzal, line 26: Have you checked that a key will always have a '+' sign?

packages/catalog-search/src/SearchSuggestions.jsx
```diff
@@ -0,0 +1,69 @@
...
+                            </div>
+                          </div>
+                        </Link>
+                      ))
+                  }
+      </div>
+      <div>
+        <div className="mb-2 mt-5 ml-2 font-weight-bold suggestions-section">
```
zamanafzal, line 34: The code for suggestion-section in course and programs looks similar. Can we make a separate compoenent and use that for both?

packages/catalog-search/src/SearchSuggestions.jsx
```diff
@@ -0,0 +1,69 @@
...
+      <div>
+        <div className="mb-2 ml-2 mt-1 font-weight-bold suggestions-section">
+          Courses
+        </div>
+        {
+                    autoCompleteHits.filter(hit => hit.content_type === 'course')
+                      .slice(0, MAX_NUM_SUGGESTIONS)
+                      .map((hit) => (
+                        <Link to={getLinkToCourse(hit)} key={hit.title} className="suggestion-item" style={{ whiteSpace: 'pre-wrap' }}>
+                          <div style={{ display: 'flex' }}>
```
zamanafzal, line 22: Please create a class for this if it is not already available in paragon?

zamanafzal (changes requested):
Overall looks good. Few changes needed.

packages/catalog-search/src/SearchSuggestions.jsx
muneebGH: this is sent by algolia as string containing html tag to distinguish the part matching search result, shouldn't it be safe?

packages/catalog-search/src/SearchSuggestions.jsx
muneebGH: yes, it is a combination of organization key and course unique identifier.

packages/catalog-search/src/SearchSuggestions.jsx
muneebGH: aggregation key is of form 'program:uuid', we need uuid only to redirect. so its this way.
copied from [here](https://github.com/openedx/frontend-app-learner-portal-enterprise/blob/c36771b0e6c30e5ae9c161e14ac9d1be6b969c33/src/components/search/SearchProgramCard.jsx#L43)

packages/catalog-search/src/SearchSuggestions.jsx
zamanafzal: Did algolia convert the text to an HTML tag? Also, please check out this link.
https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml

PR #242, opened 2022-03-30
feat: making suggested items backwards compatible and redirect func overridable

alex-sheehan-edx:
This PR:
- allows search header/suggestions to work without an index or enterprise slug
    - if index is not provided then suggestions will simply not render
    - if enterprise slug is not provided then a new bool prop `disableSuggestionRedirect=true` needs to be provided
- allows the search suggestion component's submit behavior to be overwritten with a function passed in through the search header prop `suggestionSubmitOverride` 
    - this function is given the inputs `course title` and the algolia course `hit` object
- tracks the focused element on the page and only displays suggested results when the search header is focused

**Merge checklist:**

**Post merge:**

### DIFF
--- a/packages/catalog-search/src/SearchBox.jsx
+++ b/packages/catalog-search/src/SearchBox.jsx
@@ -1,17 +1,27 @@
-import React, { useContext, useEffect, useState } from 'react';
+import React, {
+  useContext,
+  useEffect,
+  useState,
+  useCallback,
+} from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import { SearchField } from '@edx/paragon';
+import debounce from 'lodash.debounce';
 import { connectSearchBox } from 'react-instantsearch-dom';
 
 import { sendTrackEvent } from '@edx/frontend-platform/analytics';
 
 import { deleteRefinementAction, setRefinementAction } from './data/actions';
+import { useActiveElement } from './data/hooks';
 import { SearchContext } from './SearchContext';
 import {
   STYLE_VARIANTS,
   QUERY_PARAM_FOR_PAGE,
   QUERY_PARAM_FOR_SEARCH_QUERY,
+  ALGOLIA_ATTRIBUTES_TO_RETRIEVE,
+  SEARCH_BOX_CLASS_NAME,
+  DEBOUNCE_TIME_MS,
 } from './data/constants';
 import SearchSuggestions from './SearchSuggestions';
 
@@ -31,6 +41,8 @@ export const SearchBoxBase = ({
   index,
   filters,
   enterpriseSlug,
+  suggestionSubmitOverride,
+  disableSuggestionRedirect,
 }) => {
   const { dispatch, trackingName } = useContext(SearchContext);
 
@@ -54,40 +66,65 @@ export const SearchBoxBase = ({
     }
   };
 
+  const handleSuggestionSubmit = (hit) => {
+    // If an override is provided, call it with the hit
+    if (typeof suggestionSubmitOverride === 'function') {
+      suggestionSubmitOverride(hit);
+    }
+
+    // When a suggested search is clicked, set the course tile to the search query
+    setSearchQuery(hit.title);
+    // Do all the regular submit stuff
+    handleSubmit(hit.title);
+  };
+
+  const handleClickListener = () => setShowSuggestions(false);
   useEffect(() => {
-    document.addEventListener('click', () => setShowSuggestions(false));
+    document.addEventListener('click', handleClickListener);
+    // Clean up the event listener when the component unmounts-
+    // https://blog.logrocket.com/understanding-react-useeffect-cleanup-function/
+    return () => {
+      document.removeEventListener('click', handleClickListener);
+    };
   });
 
-  useEffect(() => {
-    let isMounted = true;
-    const fetchCourses = setTimeout(async () => {
-      if (searchQuery !== '' && isMounted) {
-        const { hits, nbHits } = await index.search(searchQuery, {
-          filters,
-          attributesToHighlight: ['title'],
-          attributesToRetrieve: [
-            'key',
-            'content_type',
-            'title',
-            'authoring_organizations',
-            'aggregation_key',
-            '_highlightResult',
-            'program_type',
-          ],
-        });
-        if (nbHits > 0 && isMounted) {
-          setAutocompleteHits(hits);
-          setShowSuggestions(true);
-        }
+  // Track the focused element
+  const focusedElement = useActiveElement();
+
+  // Function to be called when the user stops typing, will fetch algolia hits for query after `DEBOUNCE_TIME_MS` has
+  // elapsed
+  const debounceFunc = async (query) => {
+    // Skip suggesting anything until there's something typed
+    if (query !== '') {
+      const { hits, nbHits } = await index.search(query, {
+        filters,
+        attributesToHighlight: ['title'],
+        attributesToRetrieve: ALGOLIA_ATTRIBUTES_TO_RETRIEVE,
+      });
+      if (nbHits > 0) {
+        setAutocompleteHits(hits);
+        setShowSuggestions(true);
       } else {
+        // If there are no results of the suggested search, hide the empty suggestion component
         setShowSuggestions(false);
       }
-    }, 1000);
-    return () => {
-      isMounted = false;
-      clearTimeout(fetchCourses);
-    };
-  }, [searchQuery]);
+    // Hide the results as soon as the user removes the entire query string, instead of waiting a second
+    } else {
+      setShowSuggestions(false);
+    }
+  };
+  // Since the debounced method is called in a useEffect hook, use `useCallback` to account for repeated invoking of the
+  // method.
+  const debounceHandler = useCallback(debounce(debounceFunc, DEBOUNCE_TIME_MS), []);
+
+  useEffect(() => {
+    // If the component is provided an index and the current focused element is the input field,
+    // start the suggestion request to algolia
+    if (index !== undefined && focusedElement.classList.contains(SEARCH_BOX_CLASS_NAME)) {
+      debounceHandler(searchQuery);
+    }
+  // Retry this method if the focused element or the search query changes
+  }, [searchQuery, focusedElement]);
 
   /**
    * Handles when a search is cleared by removing the user's search query
@@ -122,7 +159,7 @@ export const SearchBoxBase = ({
         }}
       >
         <SearchField.Input
-          className="form-control-lg"
+          className={`form-control-lg ${SEARCH_BOX_CLASS_NAME}`}
           aria-labelledby="search-input-box"
           data-nr-synth-id="catalog-search-input-field"
           data-hj-whitelist
@@ -135,7 +172,9 @@ export const SearchBoxBase = ({
         <SearchSuggestions
           enterpriseSlug={enterpriseSlug}
           autoCompleteHits={autocompleteHits}
-          handleViewAllClick={() => handleSubmit(searchQuery)}
+          handleSubmit={() => handleSubmit(searchQuery)}
+          handleSuggestionClickSubmit={hit => handleSuggestionSubmit(hit)}
+          disableSuggestionRedirect={disableSuggestionRedirect}
         />
       )}
     </div>
@@ -148,9 +187,11 @@ SearchBoxBase.propTypes = {
   variant: PropTypes.oneOf([STYLE_VARIANTS.default, STYLE_VARIANTS.inverse]),
   headerTitle: PropTypes.string,
   hideTitle: PropTypes.bool,
-  index: PropTypes.shape({ search: PropTypes.func.isRequired }).isRequired,
+  index: PropTypes.shape({ search: PropTypes.func.isRequired }),
   filters: PropTypes.string,
-  enterpriseSlug: PropTypes.string.isRequired,
+  enterpriseSlug: PropTypes.string,
+  suggestionSubmitOverride: PropTypes.func,
+  disableSuggestionRedirect: PropTypes.bool,
 };
 
 SearchBoxBase.defaultProps = {
@@ -160,6 +201,10 @@ SearchBoxBase.defaultProps = {
   headerTitle: undefined,
   hideTitle: false,
   filters: '',
+  enterpriseSlug: undefined,
+  index: undefined,
+  suggestionSubmitOverride: undefined,
+  disableSuggestionRedirect: false,
 };
 
 export default connectSearchBox(SearchBoxBase);
--- a/packages/catalog-search/src/SearchHeader.jsx
+++ b/packages/catalog-search/src/SearchHeader.jsx
@@ -13,10 +13,24 @@ export const searchBoxColTestId = 'search-box-col';
 export const filtersColTestId = 'filters-col';
 
 const SearchHeader = ({
-  variant, containerSize, headerTitle, hideTitle, index, filters, enterpriseConfig: { slug },
+  variant,
+  containerSize,
+  headerTitle,
+  hideTitle,
+  index,
+  filters,
+  suggestionSubmitOverride,
+  enterpriseConfig: { slug },
+  disableSuggestionRedirect,
 }) => {
   const { refinements } = useContext(SearchContext);
-  const searchQueryFromRefinements = refinements.q;
+  let searchQueryFromRefinements;
+  // Sometimes the query is set to an array of one string instead of just the string
+  if (Array.isArray(refinements.q)) {
+    [searchQueryFromRefinements] = refinements.q;
+  } else {
+    searchQueryFromRefinements = refinements.q;
+  }
 
   return (
     <div className="bg-brand-primary">
@@ -37,6 +51,8 @@ const SearchHeader = ({
               index={index}
               filters={filters}
               enterpriseSlug={slug}
+              suggestionSubmitOverride={suggestionSubmitOverride}
+              disableSuggestionRedirect={disableSuggestionRedirect}
             />
           </Col>
           <Col
@@ -58,6 +74,10 @@ SearchHeader.defaultProps = {
   headerTitle: undefined,
   hideTitle: false,
   filters: '',
+  enterpriseConfig: { slug: undefined },
+  suggestionSubmitOverride: undefined,
+  disableSuggestionRedirect: false,
+  index: undefined,
 };
 
 SearchHeader.propTypes = {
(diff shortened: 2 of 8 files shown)


packages/catalog-search/src/data/hooks.js
```diff
@@ -73,3 +73,20 @@ export const getCatalogString = (catalogs) => {
 
   return catalogs.reduce(catalogFilterReducer, '');
 };
+
+export const useActiveElement = () => {
+  const [active, setActive] = useState(document.activeElement);
+
+  const handleFocusIn = () => {
```
binodpant, line 80: nit: 

```
const handleFocusIn = () => setActive(document.activeElement);
```

packages/catalog-search/src/SearchSuggestions.jsx
```diff
@@ -25,6 +30,9 @@ const SearchSuggestions = ({ autoCompleteHits, enterpriseSlug, handleViewAllClic
                 highlightedTitle={hit._highlightResult.title.value}
                 authoringOrganization={hit.key && hit.key.split('+')[0]}
                 title={hit.title}
+                hit={hit}
```
binodpant, line 33: upto you this one, but given we are passing `hit` here now, seems like passing in other props like authoringOrg and title that depend on hit, may be superflous ... but if it's too much refactor we can hold back on changing this

packages/catalog-search/src/SearchSuggestionItem.jsx
```diff
@@ -3,43 +3,60 @@ import { Link } from 'react-router-dom';
...
 
 SearchSuggestionItem.propTypes = {
   title: PropTypes.string.isRequired,
   url: PropTypes.string.isRequired,
   highlightedTitle: PropTypes.string.isRequired,
   authoringOrganization: PropTypes.string,
   programType: PropTypes.string,
+  handleSubmit: PropTypes.func.isRequired,
+  disableSuggestionRedirect: PropTypes.bool.isRequired,
+  hit: PropTypes.shape({
```
binodpant, line 49: yeah same thought here: given hit already has a bunch of the needed fields, it seems now the props like title and authoringOrganization feel superflous (also more props means it gets harder to test the component over time)

packages/catalog-search/src/SearchBox.jsx
```diff
@@ -135,7 +161,8 @@ export const SearchBoxBase = ({
         <SearchSuggestions
           enterpriseSlug={enterpriseSlug}
           autoCompleteHits={autocompleteHits}
-          handleViewAllClick={() => handleSubmit(searchQuery)}
+          handleSubmit={(hit) => handleSuggestionSubmit(hit.title, hit)}
```
binodpant, line 164: nit leave out the parens:

```
handleSubmit={hit => handleSuggestionSubmit(hit.title, hit)}
```

packages/catalog-search/src/SearchSuggestionItem.jsx
```diff
@@ -3,43 +3,60 @@ import { Link } from 'react-router-dom';
...
 
 const SearchSuggestionItem = ({
-  title, url, highlightedTitle, authoringOrganization, programType,
-}) => (
-  <Link to={url} key={title} className="suggestion-item">
-    <div>
-      {/* eslint-disable-next-line react/no-danger */}
-      <div dangerouslySetInnerHTML={{ __html: highlightedTitle }} />
-      {
+  title, url, highlightedTitle, authoringOrganization, programType, handleSubmit, disableSuggestionRedirect, hit,
```
binodpant, line 6: if the only use of `disableSuggestionRedirect` is to make it call `handleSubmit` then i think we can do with just one prop instea dof two.

e.g., if handleSubmit is renamed to `suggestionItemHandler` or so, you can only call it if it's defined, else do the default behavior

binodpant: like 

```
if (typeof suggestionItemHandler === 'function') { .. your new stuff }
else { old stuff }
```

packages/catalog-search/src/SearchBox.jsx
```diff
@@ -54,40 +59,61 @@ export const SearchBoxBase = ({
...
+
+    // When a suggested search is clicked, set the course tile to the search query
+    setSearchQuery(query);
+    // Do all the regular submit stuff
+    handleSubmit(query);
+  };
+
   useEffect(() => {
     document.addEventListener('click', () => setShowSuggestions(false));
   });
```
adamstankiewicz, line 76: This should likely have a clean up function to remove the event listener when the component is unmounted.

https://blog.logrocket.com/understanding-react-useeffect-cleanup-function/

packages/catalog-search/src/SearchBox.jsx
```diff
@@ -54,40 +59,61 @@ export const SearchBoxBase = ({
     }
   };
 
+  const handleSuggestionSubmit = (query, hit) => {
+    // If an override is provided, call it with the hit
+    if (suggestionSubmitOverride !== undefined) {
```
adamstankiewicz, line 64: [curious] Could `suggestionSubmitOverride` ever be `null` here? Or does passing `null` convert to `undefined` due to the default prop? 🤔

packages/catalog-search/src/SearchBox.jsx
```diff
@@ -54,40 +59,61 @@ export const SearchBoxBase = ({
...
-            'program_type',
-          ],
-        });
-        if (nbHits > 0 && isMounted) {
-          setAutocompleteHits(hits);
-          setShowSuggestions(true);
+
+    // If the component is provided an index and the current focused element is the input field,
+    // start the suggestion request to algolia
+    if (index !== undefined && focusedElement.classList.contains(SEARCH_BOX_CLASS_NAME)) {
```
adamstankiewicz, line 86: Similar question. Could `index` ever be `null` here, or if `null` is provided, does the default prop take precedence?

packages/catalog-search/src/SearchBox.jsx
```diff
@@ -54,40 +59,61 @@ export const SearchBoxBase = ({
...
+      const fetchCourses = setTimeout(async () => {
```
adamstankiewicz, line 87: [clarification] (probably a separate task) It might not make sense to use `setTimeout` here. Would this potentially be making many more queries to Algolia than we need to be making, in that on every keystroke, this `useEffect` will create a new timeout function that resolves after a second? Typically for autocomplete/typeahead, `debounce` would be used where the callback function doesn't get called until the function stops being called for a specified amount of time (e.g., don't query Algolia until 100ms _after_ the last character I type).

packages/catalog-search/src/SearchHeader.jsx
```diff
@@ -13,10 +13,24 @@ export const searchBoxColTestId = 'search-box-col';
...
+  filters,
+  suggestionSubmitOverride,
+  disableSuggestionRedirect,
+  enterpriseConfig: { slug },
 }) => {
   const { refinements } = useContext(SearchContext);
-  const searchQueryFromRefinements = refinements.q;
+  let searchQueryFromRefinements;
+  // Sometimes the query is set to an array of one string instead of just the string
+  if (refinements.q instanceof Array) {
```
adamstankiewicz, line 29: nit: alternatively `Array.isArray(refinements.q)`

packages/catalog-search/src/SearchBox.jsx
binodpant: Agreed, we should check how this operates. I think Alex is updating existing code in place but I also agree thatt debounce seems like a better pattern when doing autocomplete, in general, that is commonly done.

I am also open to the idea of doing that as a separate task if Alex think it's cleaner, but yes totally on board

binodpant: e.g., https://github.com/edx/frontend-app-admin-portal/blob/f385ac0795e264f6c5c42b7d413ea3e5d2e368eb/src/components/subscriptions/licenses/LicenseManagementTable/index.jsx

adamstankiewicz: In case it's helpful, here's some prior art on using `debounce` from `lodash.debounce`: https://github.com/openedx/frontend-app-admin-portal/blob/869afb279807f0cd1cf955314a2bcbe48d3ceda5/src/components/SubsidyRequestManagementTable/data/hooks.js#L100

adamstankiewicz: Oh, beat me to it @binodpant haha! :D Now there's 2 examples.

packages/catalog-search/src/SearchSuggestions.jsx
alex-sheehan-edx: I had the same thought when adding hit, I can definitely refactor

packages/catalog-search/src/SearchBox.jsx
alex-sheehan-edx: hmmm let me check!

alex-sheehan-edx: Hmm so, simply not providing the prop will allow things to work as expected because the default value forces it to undefined. However, you can break this by manually passing `null` in through the `suggestionSubmitOverride` prop. That should be ok though, right? Because why would someone force it to null when they could just not provide the prop?

packages/catalog-search/src/SearchBox.jsx
alex-sheehan-edx: similar situation, no `index` prop will force it to undefined. You could provide null through the prop to break things but you could also do that by providing any other junk value to the prop

packages/catalog-search/src/SearchBox.jsx
binodpant: since it's a function, I think the safest bet is to do this?

```if (typeof suggestionSubmitOverride === 'function')```

this way it will be even safer to null ,undefined or any other non functoin value

File: packages/catalog-search/src/SearchBox.jsx [hunk 11]
```diff
@@ -54,40 +61,65 @@ export const SearchBoxBase = ({
...
-        if (nbHits > 0 && isMounted) {
-          setAutocompleteHits(hits);
-          setShowSuggestions(true);
-        }
+  // Track the focused element
+  const focusedElement = useActiveElement();
+
+  // Function to be called when the user stops typing, will fetch algolia hits for query after `DEBOUNCE_TIME_MS` has
+  // elapsed
+  const debounceFunc = async (searchQuery) => {
```
binodpant, line 91: ouch, a simple fix is to just rename the searchQuery to sth else here

packages/catalog-search/src/SearchSuggestions.jsx
```diff
@@ -40,16 +44,14 @@ const SearchSuggestions = ({ autoCompleteHits, enterpriseSlug, handleViewAllClic
...
                 key={hit.title}
                 url={getLinkToProgram(hit)}
-                /* eslint-disable-next-line no-underscore-dangle */
-                highlightedTitle={hit._highlightResult.title.value}
-                authoringOrganization={hit.authoring_organizations.shift()?.key}
-                title={hit.title}
-                programType={hit.program_type}
+                hit={hit}
+                suggestionItemHandler={handleSubmit}
+                disableSuggestionRedirect={disableSuggestionRedirect}
```
binodpant, line 49: hmm looks like you don't actually use this anymore in SearchSuggestionItem? Can it be removed?

packages/catalog-search/src/SearchBox.jsx
```diff
@@ -135,7 +171,7 @@ export const SearchBoxBase = ({
         <SearchSuggestions
           enterpriseSlug={enterpriseSlug}
           autoCompleteHits={autocompleteHits}
-          handleViewAllClick={() => handleSubmit(searchQuery)}
+          handleSubmit={hit => handleSuggestionSubmit(hit.title, hit)}
```
muneebGH, line 174: when this `handleSubmit` will be called by the View all results button click then there wont be any hit passed? the desired actions would be performed in that case? is this expected and taken care of? what learner portal expects is that when view all is clicked then **whatever the user have typed** will be the new query and all search results below will be updated.

when clicked on `SearchSuggestionItem` according to the logic if we provide override then that would be called along with setting the search query in search box to the title of the item we just clicked, which should happen? because the title isnt the query and we may not want to set the text in search box to the title of item we just hit? is that it or am i missing something?

packages/catalog-search/src/SearchBox.jsx
```diff
@@ -54,40 +65,65 @@ export const SearchBoxBase = ({
     }
   };
 
+  const handleSuggestionSubmit = (query, hit) => {
```
muneebGH, line 68: the understanding i got from this function is that when we wont provide the override then clicking on the `SearchSuggestionItem` wont be affected on learner portal, is that right?


and about the click on view all results i have a concern which is mentioned in another comment.

muneebGH (changes requested):

packages/catalog-search/src/SearchBox.jsx
alex-sheehan-edx: this method is provided to the suggestion items and is only invoked when `disableSuggestionRedirect` is true. Instead of redirecting the user to the url (in the learner portal's case, the enterprise course page), it defaults to whatever functionality was provided through this method.

packages/catalog-search/src/SearchBox.jsx
alex-sheehan-edx: clicking view all will set the original search query, clicking the individual row will set the query as the title of the clicked row.

packages/catalog-search/src/tests/SearchBox.test.jsx
```diff
@@ -65,48 +65,42 @@ describe('<SearchBox />', () => {
...
-        attributesToRetrieve: [
-          'key',
-          'content_type',
-          'title',
-          'authoring_organizations',
-          'aggregation_key',
-          '_highlightResult',
-          'program_type',
-        ],
+        attributesToRetrieve: ['*'],
```
binodpant, line 79: Is this change related to autocomplete? Do we now get all attributes instead of the limited list before?

packages/catalog-search/src/SearchBox.jsx
```diff
@@ -31,6 +41,8 @@ export const SearchBoxBase = ({
   index,
   filters,
   enterpriseSlug,
+  suggestionSubmitOverride,
+  disableSuggestionRedirect,
```
binodpant, line 45: hmm looks like disableSuggestionRedirect is a pass through prop that is really only used by other components. In that case do you thin it makes sense to set it as 

```
disableSuggestionRedirect = typeof suggestionSubmitOverride === 'function'
```

then pass it along to the other components? This way you will end up with only one new prop here instead of two

Just a thought to see if we can further simplify things.

binodpant: also may be easier for the authors using this component to think of one concept as one prop, rather than two...

packages/catalog-search/src/SearchSuggestionItem.jsx
```diff
@@ -3,43 +3,56 @@ import { Link } from 'react-router-dom';
...
-      {/* eslint-disable-next-line react/no-danger */}
-      <div dangerouslySetInnerHTML={{ __html: highlightedTitle }} />
-      {
+  url, suggestionItemHandler, hit, disableSuggestionRedirect,
+}) => {
+  const authoringOrganization = hit.key && hit.key.split('+')[0];
+  // If the disable redirect bool is provided, prevent the redirect from happening and instead call the provided submit
+  // handler
+  const handleLinkDisable = (e) => {
+    if (disableSuggestionRedirect) {
```
binodpant, line 12: I think we may be able to remove the disableSuggestionRedirect flag and instead use the validity of the suggestionItemHandler to replace this check? What do you think?

packages/catalog-search/src/tests/SearchBox.test.jsx
alex-sheehan-edx: this change is specifically for displaying course info modals in public catalog using the hit information provided from frontend-enterprise

packages/catalog-search/src/SearchSuggestionItem.jsx
alex-sheehan-edx: talked offline and decided to resolve this in a fast follow. Ticket TBD

packages/catalog-search/src/SearchBox.jsx
```diff
@@ -54,40 +66,65 @@ export const SearchBoxBase = ({
...
   };
 
+  const handleSuggestionSubmit = (hit) => {
+    // If an override is provided, call it with the hit
+    if (typeof suggestionSubmitOverride === 'function') {
+      suggestionSubmitOverride(hit);
+    }
+
+    // When a suggested search is clicked, set the course tile to the search query
+    setSearchQuery(hit.title);
```
muneebGH, line 76: My only concern now is that after we have split View all click handler, is there any need for these lines?
```
setSearchQuery(hit.title);
handleSubmit(hit.title);
```

because this function will only be called when someone clicks some item from suggestion and redirect is disabled. So this function wont be called on learner portal but only on explore catalog, so do you guys want this behaviour that when user clicks on suggestion then along with you callback getting called, you want all the search results to be refined according to the title of item user just clicked? and searchbox to contain whatever the title of item user clicked? if yes then all good.

muneebGH (approved):
LGTM, Nice work @alex-sheehan-edx 🎉 
just one question posted in comments

packages/catalog-search/src/SearchBox.jsx
muneebGH: oky cool, thanks @alex-sheehan-edx @binodpant

packages/catalog-search/src/SearchBox.jsx
alex-sheehan-edx: Hmm, I was trying to preserve the behavior of the original learner portal which is why I excluded those from the handle submit method. The thinking is that there are 3 functionalities currently supported (and @binodpant had some great suggestions as to how we might be able to get ourselves to just two or even one) Currently the search header/box suggestions work in such a way that:
- if  the `disableSuggestionRedirect`prop is provided, then a base behavior of setting the facet/input value to the suggested search title is used instead of redirecting to the course page
- if `suggestionSubmitOverride` is provided along with `disableSuggestionRedirect` then an additional custom behavior is applied to the onclick of the suggestion along with the new base behavior
- if no new props are provided then the behavior reflects current day learner portal implementation (we might rework this to the above `additional behavior` functionality.

(also side note, not giving the index prop will disable suggestions all together)

binodpant (approved):

PR #249, opened 2022-04-05
fix: resolves many peer dependency issues

adamstankiewicz:
**Merge checklist:**

**Post merge:**


package.json
```diff
@@ -20,13 +20,7 @@
...
   "devDependencies": {
     "@commitlint/config-conventional": "13.2.0",
-    "@edx/frontend-platform": "1.9.6",
-    "@edx/paragon": "14.5.0",
     "husky": "6.0.0",
-    "lerna": "4.0.0",
-    "react": "16.14.0",
-    "react-dom": "16.14.0",
-    "react-router-dom": "5.2.0",
-    "react-test-renderer": "16.14.0"
```
adamstankiewicz, line 30: These dependencies were previously installed at the root `node_modules`, as when importing these dependencies in `catalog-search`, `utils`, `logistration`, etc. would resolve correctly to the parent `node_modules`, instead of local to that package.

However, this causes many peer dependency issues since they are not explicitly defined in the package.json files of packages like `catalog-search`, etc.

So, instead, we will explicitly move these dependencies to each individual package, as needed, at the expense of no longer having a single source of truth version of `react`, `react-dom`, etc. throughout the monorepo.

packages/catalog-search/package.json
```diff
@@ -6,9 +6,6 @@
     "type": "git",
     "url": "git+https://github.com/edx/frontend-enterprise.git"
   },
-  "browserslist": [
-    "extends @edx/browserslist-config"
-  ],
```
adamstankiewicz, line 11: This was added prematurely. Removing because migrating to this shared `@edx/browserslist-config` is blocked on Node 16 upgrade.

packages/catalog-search/package.json
```diff
@@ -43,28 +40,35 @@
...
+    "@edx/paragon": "14.5.0",
     "@fortawesome/free-solid-svg-icons": "5.8.1",
     "@fortawesome/react-fontawesome": "0.1.4",
     "@testing-library/jest-dom": "5.11.6",
     "@testing-library/react": "11.2.6",
     "@testing-library/react-hooks": "3.4.2",
     "@testing-library/user-event": "13.1.8",
-    "npm-watch": "^0.9.0",
-    "react-instantsearch-dom": "6.8.3"
+    "algoliasearch": "4.13.0",
```
adamstankiewicz, line 56: Required to resolve a peer dependency warning.

packages/logistration/src/LoginRedirect.jsx
```diff
@@ -14,15 +14,15 @@ import { getProxyLoginUrl } from './utils';
  */
 export default function LoginRedirect({ children, loadingDisplay: LoadingDisplay }) {
   const user = getAuthenticatedUser();
+  const {
+    enterpriseSlug,
+    enterpriseCustomerInviteKey,
+  } = useParams();
```
adamstankiewicz, line 20: Hooks must be rendered in same order, regardless of conditions.

PR #250, opened 2022-04-05
Node 16 upgrade and peer dependency cleanup

adamstankiewicz:
## Description

* Moves from Node 12 ➡️ Node 16, as **we now rely on NPM workspaces!** 🚀 
* Resolves the majority of our peer dependency issues, at least the ones that block an `npm install` on Node 16+.
  * As a result, most dependencies installed in the root `package.json` were instead distributed to each individual package as necessary. The tradeoff is we lose having a single source-of-truth of common "global" dependencies like `@edx/frontend-platform`, but this is OK and was a premature optimization (i.e., we only have 3 packages).
* Updates `.nvmrc` to include `16` :tada: 
* Moves away from relying on `lerna bootstrap`, which is a command that is a little magical ✨: 
  * It does a combination of installing packages, creating/updating `package-lock.json`, and creating symlinks between packages in this monorepo to allow importing by the module name, e.g. `import { useEnterpriseConfig } from '@edx/frontend-enterprise-utils';`.
  * The symlinking behavior was causing discrepancies between local environments and CI, so instead, we will now rely on running `npm install` ourselves via `lerna exec -- npm install` to avoid this symlinking wizardry altogether 🧙 
  * ~I think this also may be a step towards using NPM `workspaces` in the future?~ Adopts NPM workspaces! 😄 

## Testing

Using `@edx/frontend-enterprise-utils` via a `module.config.js` override in `frontend-component-header-edx` seemingly led to good results ([related PR](https://github.com/edx/frontend-component-header-edx/pull/231)):

image

Additionally, all 3 packages (`@edx/frontend-enterprise-utils`, `@edx/frontend-enterprise-catalog-search`, and `@edx/frontend-enterprise-logistration`) are used by `frontend-app-admin-portal`. In a test via `module.config.js`, the application seems to run no problem 💪 

image

## The Other stuff

**Merge checklist:**
- [x] Evaluate how your changes will impact existing consumers (e.g., `frontend-app-learner-portal-enterprise`, `frontend-app-admin-portal`, and `frontend-app-enterprise-public-catalog`). Will consumers safely be able to upgrade to this change without any breaking changes?
- [x] Ensure your commit message follows the semantic-release conventional commit message format. If your changes include a breaking change, ensure your commit message is explicitly marked as a `BREAKING CHANGE` so the NPM package is released as such.
- [x] Once CI is passing, verify the package versions that Lerna will increment to in the Github Action CI workflow logs.

**Post merge:**


Jawayria (approved):

.github/workflows/ci.yml
```diff
@@ -3,44 +3,47 @@ name: CI
...
+    # lerna expects to be authenticated for publishing to NPM. This step will fail CI if NPM is not
+    # authenticated, even though this build does _not_ attempt to publish, as an extra check before merge
+    # that Lerna is set to publish.
+    - name: Check NPM authentication
+      run: |
+        echo "//registry.npmjs.org/:_authToken=${{ secrets.SEMANTIC_RELEASE_NPM_TOKEN }}" >> .npmrc
+        npm whoami
+    - name: Install and Setup Dependencies
+      run: |
+        npm install
```
binodpant, line 33: interesting we need to run two install commands?

binodpant (approved):

.github/workflows/ci.yml
adamstankiewicz: The first is installing node modules at the root directory (e.g, to install Lerna itself).

The second is to run npm install in all packages. So technically it's 4 installs 🤪

long74100 (approved):

.github/workflows/ci.yml
binodpant: ah I see, makes sense now, thanks!

binodpant: more the (merrier) 🤣

.github/workflows/ci.yml
```diff
@@ -1,46 +1,42 @@
...
+      uses: actions/checkout@v2
+      with:
+        # pulls all commits (needed for lerna / semantic release to correctly version)
+        fetch-depth: 0
+    # pulls all tags (needed for lerna / semantic release to correctly version)
+    - run: git fetch --depth=1 origin +refs/tags/*:refs/tags/*
+    - name: Setup Nodejs
+      uses: actions/setup-node@v3
+      with:
+        node-version: 16
```
adamstankiewicz, line 20: This repository is now relying on NPM workspaces. As such, it is not compatible with Node 12 or Node 14, as NPM workspaces require `npm` v7+.

.github/workflows/ci.yml
```diff
@@ -1,46 +1,42 @@
...
+    - name: Lint
+      run: lerna run lint
+    - name: Test
+      run: lerna run test
+    - name: Coverage Report
+      uses: codecov/codecov-action@v2
+    - name: Preview Updated Versions (dry run)
+      run: |
+        git switch -c "actual-branch-not-detached-head"
+        lerna version --allow-branch actual-branch-not-detached-head --no-git-tag-version --no-changelog --no-push --yes
```
adamstankiewicz, line 42: By changing the `on` from "push" to "pull_request" ensures this workflow is only run when necessary, but in doing so, the checkout is in a detached `HEAD` state, so we need to force switch to an attached branch for the version dry-run command to work.

.github/workflows/ci.yml
```diff
@@ -1,46 +1,43 @@
...
-        id: extract_branch
-      - name: Preview Updated Versions (dry run)
-        run: lerna version --allow-branch ${{ steps.extract_branch.outputs.branch }} --no-git-tag-version --no-changelog --no-push --yes
+    - name: Checkout
+      uses: actions/checkout@v2
+      with:
+        # pulls all commits (needed for lerna / semantic release to correctly version)
+        fetch-depth: 0
+    # pulls all tags (needed for lerna / semantic release to correctly version)
+    - run: git fetch --depth=1 origin +refs/tags/*:refs/tags/*
```
macdiesel, line 16: I'm curious as to why this run directive is the start of it's own sequence and not part of the "Checkout" sequence?

macdiesel (approved):

PR #257, opened 2022-05-11
feat: Add package-lock file version check

Jawayria:
lockfileVersion for the package-lock.json generated by NPM6 was 1 but now as we've moved to Node16 and NPM8, package-lock.json should be generated using NPM8. lockfileVersion for package-lock.json generated using NPM8 is 2. So here we're going to check the lockfileVersion in package-lock.json

### DIFF
--- a/.github/workflows/lockfileversion-check.yml
+++ b/.github/workflows/lockfileversion-check.yml
@@ -1,0 +1,13 @@
+#check package-lock file version
+
+name: lockfileVersion check
+
+on:
+  push:
+    branches:
+    - master
+  pull_request:
+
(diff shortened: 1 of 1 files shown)


.github/workflows/lockfileversion-check.yml
```diff
@@ -0,0 +1,13 @@
+#check package-lock file version
+
+name: lockfileVersion check
```
aht007, line 3: Don't forget to update the name here before merging

aht007 (approved):

UsamaSadiq (approved):

PR #259, opened 2022-06-02
fix: Use media query to resize for mobile in SearchFilters

iloveagent57:
ENT-5800.  Also installs react-responsive as a peer dependency.

Something about `useWindowSize()` from paragon isn't working quite right with the edx.org theme, so our original solution for https://2u-internal.atlassian.net/browse/ENT-5800 isn't working.  From Adam:
> I think the issue is that useWindowSize is including the extra width of the overflow facets in the total width. for example, at a window width of 1190px but useWindowSize says it's at 1249px, which is still above the breakpoint.

Note that there's still a little overrun from 1200px < screen width < ~1350px, but it's far better than before and I don't want to squish into the mobile view for reasonably-sized desktop browser windows by switching to the XL breakpoint.

media-query

**Merge checklist:**
- [x] Evaluate how your changes will impact existing consumers (e.g., `frontend-app-learner-portal-enterprise`, `frontend-app-admin-portal`, and `frontend-app-enterprise-public-catalog`). Will consumers safely be able to upgrade to this change without any breaking changes?
- [x] Ensure your commit message follows the semantic-release conventional commit message format. If your changes include a breaking change, ensure your commit message is explicitly marked as a `BREAKING CHANGE` so the NPM package is released as such.
- [x] Once CI is passing, verify the package versions that Lerna will increment to in the Github Action CI workflow logs.

**Post merge:**

### DIFF
--- a/packages/catalog-search/package.json
+++ b/packages/catalog-search/package.json
@@ -45,7 +45,7 @@
     "@edx/browserslist-config": "1.0.0",
     "@edx/frontend-build": "9.1.4",
     "@edx/frontend-platform": "1.15.6",
-    "@edx/paragon": "14.5.0",
+    "@edx/paragon": "^19.25.1",
     "@fortawesome/free-solid-svg-icons": "5.8.1",
     "@fortawesome/react-fontawesome": "0.1.4",
     "@testing-library/jest-dom": "5.11.6",
--- a/packages/catalog-search/src/SearchFilters.jsx
+++ b/packages/catalog-search/src/SearchFilters.jsx
@@ -1,6 +1,6 @@
 import React, { useMemo, useContext } from 'react';
 import PropTypes from 'prop-types';
-import { breakpoints, useWindowSize } from '@edx/paragon';
+import { breakpoints, MediaQuery } from '@edx/paragon';
 
 import FacetListRefinement from './FacetListRefinement';
 import FacetListBase from './FacetListBase';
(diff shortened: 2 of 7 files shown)


packages/catalog-search/package.json
```diff
@@ -66,6 +66,7 @@
     "react": "^16.12.0",
     "react-dom": "^16.12.0",
     "react-instantsearch-dom": "^6.8.3",
+    "react-responsive": "^8.2.0",
```
adamstankiewicz, line 69: `MediaQuery` is a pass-thru component from `@edx/paragon` already, so we shouldn't need to explicitly install `react-responsive` here.

packages/catalog-search/src/SearchFilters.jsx
```diff
@@ -85,18 +81,20 @@ const SearchFilters = ({ variant }) => {
...
   return (
     <>
-      {showMobileMenu ? (
+      <MediaQuery maxWidth={breakpoints.large.maxWidth}>
         <MobileFilterMenu className="mb-3">
           {searchFacets}
         </MobileFilterMenu>
-      ) : (
+      </MediaQuery>
+
```
adamstankiewicz, line 89: nit: unnecessary blank space

adamstankiewicz (approved):

packages/catalog-search/package.json
iloveagent57: fwiw, the linter did insist that I install it in `dependencies`:
```  4:24  error  'react-responsive' should be listed in the project's dependencies. Run 'npm i -S react-responsive' to add it  import/no-extraneous-dependencies```

PR #271, opened 2022-08-23
feat: add initial Toasts and ToastProvider component to package

brobro10000:
https://2u-internal.atlassian.net/browse/ENT-5504

Implemented package integrating Toasts, and ToastsProvider. Currently pending PRs on the following repos:
[frontend-app-learner-portal-enterprise](https://github.com/openedx/frontend-app-learner-portal-enterprise/pull/605)
[frontend-app-admin-portal](https://github.com/openedx/frontend-app-admin-portal/pull/846)

Requires further unit testing, posting PR as first look to improve implementation upon feedback.

**Merge checklist:**

**Post merge:**


openedx-webhooks:
Thanks for the pull request, @brobro10000!  Please note that it may take us up to several weeks or months to complete a review and merge your PR.

Feel free to add as much of the following information to the ticket as you can:
- supporting documentation
- [Open edX discussion forum threads](https://discuss.openedx.org/)
- timeline information ("this must be merged by XX date", and why that is)
- partner information ("this is a course on edx.org")
- any other information that can help Product understand the context for the PR

All technical communication about the code itself will be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).


Please let us know once your PR is ready for our review and all tests are green.


:warning: We can't start reviewing your pull request until you've submitted a     [signed contributor agreement](https://openedx.org/cla)     or indicated your institutional affiliation. Please see the [CONTRIBUTING](https://github.com/edx/edx-platform/blob/master/CONTRIBUTING.rst)     file for more information.     If you've signed an agreement in the past, you may need to re-sign.  See     [The New Home of the Open edX Codebase](https://open.edx.org/blog/the-new-home-of-the-open-edx-codebase/)     for details.

packages/toasts/src/ToastsProvider.jsx
```diff
@@ -0,0 +1,39 @@
...
+    const index = toasts.findIndex(toast => toast.id === id);
+    setToasts((prevToasts) => {
+      const newToasts = [...prevToasts];
+      newToasts.splice(index, 1);
+      return newToasts;
+    });
+  };
+
+  return (
+    <ToastsContext.Provider value={{ toasts, addToast, removeToast }}>
```
adamstankiewicz, line 29: Despite the original implementation not currently being implemented as such, typically context values need to be memoized (with `useMemo`) so that when its parent component re-renders, the object passed into `value` prop doesn't get re-computed each re-render.

Whenever the context value changes, all the children passed to this component re-renders as well. So, the memoization of the context value prevents unintended re-rendering of the children

packages/toasts/src/Toasts.jsx
```diff
@@ -0,0 +1,19 @@
...
+
+import { ToastsContext } from './ToastsProvider';
+
+const Toasts = () => {
+  const { toasts, removeToast } = useContext(ToastsContext);
+  return toasts.map(({ id, message }) => (
+    <Toast
+      key={id}
+      onClose={() => removeToast(id)}
+      show
```
adamstankiewicz, line 12: So, yeah... haha. This is exactly how it is currently implemented in original implementation but after a closer look, I realized I think we actually may be using `Toast` from Paragon incorrectly...

As is, we are rendering `Toast` immediately when we want it to be displayed via `addToast` in the MFE code, by forcing `show={true}` right away. The issue with this is that we actually lose the CSS animation of toasts fading in/out with this approach, unfortunately 😢 

Instead, I believe we may actually want to render the `Toast` component even before its officially "visible", so that the parent container element Paragon has a chance to render first before any `Toasts` are visible. That way, when a `Toast` flips from `show={false}` to `show={true}`, it can animate properly.

By waiting to render a `Toast` only when we want it to be visible (where `show` always is true) means the toast container is not in the DOM initially. By the time a `Toast` is rendered when `addToast` is called, it's both adding the toast container _and_ the individual toast itself simultaneously, which doesn't give the browser an opportunity to do its CSS animations for the now visible toast.

Check out the fade in/out animation effect on the [Paragon documentation site](https://paragon-openedx.netlify.app/components/toast/#basic-usage). You'll notice in our enterprise learner/admin portal MFEs, wherever we call `addToast` in its current implementation, there is no animation.


It's definitely an open question still in terms of how (or whether) we might want to address this, though... 

@brobro10000 Sorry, throwing a lot of info at you about how the original implementation we're basing this PR on wasn't solid to begin with... would love to get your take on this 😉

packages/toasts/src/ToastsProvider.jsx
```diff
@@ -0,0 +1,39 @@
+import React, { createContext, useState } from 'react';
+import PropTypes from 'prop-types';
+
+export const ToastsContext = createContext();
+
+const ToastsProvider = ({ children }) => {
+  const [toasts, setToasts] = useState([]);
+
+  const addToast = (message) => {
```
adamstankiewicz, line 9: Right now, `addToast` largely expects a string message, though, technically, a React node could be passed as well (since the `message` gets rendered in the `Toast` children). It might be worth adding a JSDoc docstring above the function to document data types the `message` argument can be.

I also wonder if there's a better function argument name than `message` since I feel `message` kind of implies a string.

adamstankiewicz: FWIW the current implementation that we're pulling this from also limits the flexibility of how `Toast` can be used. For example, `Toast` can be optionally customized to include a button/link (see Paragon docs [here](https://paragon-openedx.netlify.app/components/toast/#with-button) and [here](https://paragon-openedx.netlify.app/components/toast/#with-link)) via an `action` prop but the `addToast` examples we have today wouldn't support customizing the `action` prop on a case-by-case basis.

I wonder if there's a possible way to make this functionality more flexible to support that as well while we're in here making changes 🤔

packages/toasts/src/index.js
```diff
@@ -0,0 +1,6 @@
+// eslint-disable-next-line import/prefer-default-export
```
adamstankiewicz, line 1: nit: Is this eslint comment still relevant with multiple exports in this file now?

packages/toasts/src/ToastsProvider.jsx
```diff
@@ -0,0 +1,57 @@
...
+   * onClick: function - **Optional**. The function to be called when the action button is clicked.
+   * ** action must contain either href or onClick property **.
+   * The functionality mimics the functionality of the '@edx/paragon' Toast component.
+   * See {@link https://paragon-openedx.netlify.app/components/toast/#home}
+   */
+  const addToast = useCallback((message, actionObj = null, delayLength = 5000) => {
+    setToasts(prevToasts => [
+      ...prevToasts,
+      {
+        id: getRandom(0, 1000),
```
long74100, line 27: [Question]: Why are we using a random id here instead of the length of toasts? The id here is not guaranteed to be unique.

packages/toasts/src/en.json
```diff
@@ -0,0 +1,3 @@
+{
+    "DEFAULT_MESSAGE": "Butter"
```
long74100, line 2: I don't think this file is needed?

packages/toasts/src/ToastsProvider.test.jsx
```diff
@@ -0,0 +1,53 @@
...
+const MockButtons = () => {
+  const { addToast, removeToast } = useContext(ToastsContext);
+  return (
+    <>
+      <Toasts />
+      <button type="button" onClick={() => addToast('Butter')}>Add</button>
+      <button type="button" onClick={() => removeToast(0)}>Remove</button>
+    </>
+  );
+};
```
long74100, line 22: Let's consolidate these two components into one wrapper component. maybe something like
```
const Wrapper = () => 
  <IntlProvider>
    <ToastsProvider>
        <Toasts />
        <button type="button" onClick={() => {
          addToast('Butter');
          addToast('Github', { label: 'Test', href: 'https://github.com/' })
        }}>Add</button>
        <button type="button" onClick={() => removeToast(0)}>Remove</button>
    </ToastsProvider>
  </IntlProvider>
```

brobro10000: For this one, the test kept throwing an error regarding React.children.only was expecting a single child. Further looking into this error concluded it had to do with testing a button

packages/toasts/src/en.json
brobro10000: I got an error with the international provider requesting a default message. I believe it should only be on the test environment, but it returned a test error when no default message for language was given.

PR #288, opened 2023-02-13
feat: adding exec ed compatibility to suggested search

alex-sheehan-edx:
https://2u-internal.atlassian.net/browse/ENT-6670

image

**Merge checklist:**

**Post merge:**

### DIFF
--- a/packages/catalog-search/src/SearchSuggestions.jsx
+++ b/packages/catalog-search/src/SearchSuggestions.jsx
@@ -14,44 +14,76 @@ const SearchSuggestions = ({
   const getLinkToCourse = (course) => `/${enterpriseSlug}/course/${course.key}`;
   const getLinkToProgram = (program) => `/${enterpriseSlug}/program/${program.aggregation_key.split(':').pop()}`;
 
+  const courses = [];
+  const programs = [];
+  const execEdCourses = [];
+  autoCompleteHits.forEach((hit) => {
+    const { learning_type: learningType } = hit;
+    if (learningType === 'course') { courses.push(hit); }
+    if (learningType === 'program') { programs.push(hit); }
+    if (learningType === 'Executive Education') { execEdCourses.push(hit); }
+  });
   return (
     <div className="suggestions" data-testid="suggestions">
-      <div>
-        <div className="mb-2 ml-2 mt-1 font-weight-bold suggestions-section">
-          Courses
+      {courses.length > 0 && (
+        <div>
+          <div className="mb-2 ml-2 mt-1 font-weight-bold suggestions-section">
+            Courses
+          </div>
+          {
+            courses.slice(0, MAX_NUM_SUGGESTIONS)
+              .map((hit) => (
+                <SearchSuggestionItem
+                  key={hit.title}
+                  url={getLinkToCourse(hit)}
+                  hit={hit}
+                  disableSuggestionRedirect={disableSuggestionRedirect}
+                  suggestionItemHandler={handleSuggestionClickSubmit}
+                />
+              ))
+          }
         </div>
-        {
-          autoCompleteHits.filter(hit => hit.content_type === 'course')
-            .slice(0, MAX_NUM_SUGGESTIONS)
-            .map((hit) => (
-              <SearchSuggestionItem
-                key={hit.title}
-                url={getLinkToCourse(hit)}
-                hit={hit}
-                disableSuggestionRedirect={disableSuggestionRedirect}
-                suggestionItemHandler={handleSuggestionClickSubmit}
-              />
-            ))
-        }
-      </div>
-      <div>
-        <div className="mb-2 mt-5 ml-2 font-weight-bold suggestions-section">
-          Programs
+      )}
+      {programs.length > 0 && (
+        <div>
+          <div className="mb-2 mt-5 ml-2 font-weight-bold suggestions-section">
+            Programs
+          </div>
+          {
+            programs.slice(0, MAX_NUM_SUGGESTIONS)
+              .map((hit) => (
+                <SearchSuggestionItem
+                  key={hit.title}
+                  url={getLinkToProgram(hit)}
+                  hit={hit}
+                  disableSuggestionRedirect={disableSuggestionRedirect}
+                  suggestionItemHandler={handleSuggestionClickSubmit}
+                />
+              ))
+          }
         </div>
-        {
-          autoCompleteHits.filter(hit => hit.content_type === 'program')
-            .slice(0, MAX_NUM_SUGGESTIONS)
-            .map((hit) => (
-              <SearchSuggestionItem
-                key={hit.title}
-                url={getLinkToProgram(hit)}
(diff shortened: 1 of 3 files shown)


packages/catalog-search/src/tests/SearchSuggestions.test.jsx
```diff
@@ -67,7 +67,6 @@ describe('<SeachSuggestions />', () => {
       autoCompleteHits={fakeSuggestionsDataEmptyAuthoringOrgs.hits}
       handleSubmit={handleSubmit}
     />);
-    expect(screen.getByText('Courses')).not.toBeNull();
```
alex-sheehan-edx, line 70: made the decision that we shouldn't have an empty half container with title and no results if no results are present

packages/catalog-search/src/SearchSuggestions.jsx
```diff
@@ -14,15 +14,18 @@ const SearchSuggestions = ({
   const getLinkToCourse = (course) => `/${enterpriseSlug}/course/${course.key}`;
   const getLinkToProgram = (program) => `/${enterpriseSlug}/program/${program.aggregation_key.split(':').pop()}`;
 
+  const courses = autoCompleteHits.filter(hit => hit.learning_type === 'course');
+  const programs = autoCompleteHits.filter(hit => hit.learning_type === 'program');
+  const execEdCourses = autoCompleteHits.filter(hit => hit.learning_type === 'Executive Education');
```
adamstankiewicz, line 19: It may be worth trying to refactor this to only do the iteration over `autoCompleteHits` once to compute the `courses`, `programs`, and `execEdCourses`, e.g.:

```jsx
const courses = [];
const programs = [];
const execEdCourses = [];
autoCompleteHits.forEach((hit) => {
  const { learning_type: learningType } = hit;

  if (hit.learner_type === 'course') { courses.push(hit); }
  if (hit.learner_type === 'program') { programs.push(hit); }
  if (hit.learner_type === 'Executive Education') { execEdCourses.push(hit); }
}));
```

adamstankiewicz, line 19: [nit/curious] Kind of surprised `learning_type` comes back as a string like "Executive Education" vs. `executive_education_2u` or `executive-education-2u` or something like that?

packages/catalog-search/src/SearchSuggestions.jsx
```diff
@@ -14,15 +14,18 @@ const SearchSuggestions = ({
...
   return (
     <div className="suggestions" data-testid="suggestions">
+      { courses.length > 0 && (
```
adamstankiewicz, line 22: nit: extra space after the `{`

packages/catalog-search/src/SearchSuggestions.jsx
adamstankiewicz, line 19: [clarification] Are we intentionally not handling `pathways` as a `learning_type` here?

image

packages/catalog-search/src/SearchSuggestions.jsx
```diff
@@ -52,6 +56,28 @@ const SearchSuggestions = ({
...
+              <SearchSuggestionItem
+                key={hit.title}
+                url={getLinkToCourse(hit)}
+                hit={hit}
+                disableSuggestionRedirect={disableSuggestionRedirect}
+                suggestionItemHandler={handleSuggestionClickSubmit}
+              />
+            ))
+        }
+      </div>
```
adamstankiewicz, line 79: nit: extra indent needed/preferred

packages/catalog-search/src/SearchSuggestions.jsx
```diff
@@ -52,6 +56,28 @@ const SearchSuggestions = ({
             ))
         }
       </div>
+      )}
+      {/* Currently (Feb 2023) it is not possible to redirect to the learner portal for exec ed content so only display
+       if the redirect is disabled */}
```
adamstankiewicz, line 61: [curious] Is coming back to update this redirect logic once Learner portal can display exec ed content already ticketed somewhere?

packages/catalog-search/src/SearchSuggestions.jsx
```diff
@@ -52,6 +56,28 @@ const SearchSuggestions = ({
...
+      <div>
+        <div className="mb-2 mt-5 ml-2 font-weight-bold suggestions-section">
+          Executive Education
+        </div>
+        {
+          execEdCourses.slice(0, MAX_NUM_SUGGESTIONS)
+            .map((hit) => (
+              <SearchSuggestionItem
+                key={hit.title}
+                url={getLinkToCourse(hit)}
```
adamstankiewicz, line 72: [inform] Through https://2u-internal.atlassian.net/browse/ENT-6966, in order to support the co-branded header with the GetSmarter logo on learner portal course pages (i.e., when learner is viewing an Exec Ed course, show both customer logo & GetSmarter logo side by side), we've opted to create a new route to handle non-OCM course types, e.g.: `https://enterprise.edx.org/<slug>/executive-education-2u/course/<course_key>` which means `getLinkToCourse` ([source](https://github.com/openedx/frontend-enterprise/blob/b31a1ff73b27ad5047d1bb4a22d228498159709b/packages/catalog-search/src/SearchSuggestions.jsx#L14)) will likely need to support course types. The course pages are otherwise identical except for the co-branded header.

We do have logic in place where if the learner lands on the OCM course page route for an Exec Ed course, we redirect them to the route with the `courseType` in it so the logo appears.

[context] The rationale for this decision about course type specific course about pages is that we want to show the co-branded logos with GetSmarter immediately on page load so that the GetSmarter logo doesn't just randomly appear in the header after the course metadata from course-discovery API finishes loading. This is a shortcoming in our MFE architecture, where it'd be _really_ nice if we could fetch the prerequisite data needed for an MFE route on the server instead of at runtime, after the entire JS bundle gets loaded/processed by the browser.

packages/catalog-search/src/SearchSuggestions.jsx
alex-sheehan-edx: that was done intentionally to make the learning type a readable facet for the explore catalog

alex-sheehan-edx: Only because it doesn't really apply to my use case

packages/catalog-search/src/SearchSuggestions.jsx
alex-sheehan-edx: it is not, but I can do it

adamstankiewicz: Thanks! Re-posting here: https://2u-internal.atlassian.net/browse/ENT-7171

packages/catalog-search/src/SearchSuggestions.jsx
alex-sheehan-edx: talked offline, pathways futures are tbd and don't really apply to my use case so we're holding off

PR #289, opened 2023-02-14
Attempt to fixup Lerna package publishing automation.

macdiesel:


brobro10000 (approved):

.github/workflows/release.yml
```diff
@@ -43,3 +43,19 @@ jobs:
         run: lerna version --yes
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Generate Lerna Versions PR
+        id: vpr
+        uses: peter-evans/create-pull-request@v4
```
alex-sheehan-edx, line 48: curious what this means- why keep peter-evans?

alex-sheehan-edx (approved):
one question otherwise looks good

lerna.json
```diff
@@ -5,8 +5,8 @@
   "version": "independent",
   "command": {
     "publish": {
-      "allowBranch": "master",
-      "message": "chore(release): publish :tada:",
+      "allowBranch": "automation/lerna",
+      "message": "chore(release): Publish updated package versions to NPM :tada:",
```
adamstankiewicz, line 9: nit: This commit isn't so much about publishing direct to NPM; it's more about ensuring the package.json and CHANGELOG file(s) are updated so that when the "Publish from package.json" Github Action workflow is manually run, Lerna publishes whatever is currently on `master` at the version(s) specified in the package.json file(s) for each package.

.github/workflows/release.yml
```diff
@@ -43,3 +43,19 @@ jobs:
...
+        with:
+            token: ${{ secrets.GITHUB_TOKEN }}
+            title: Publish Packages - Lerna
+            body: |
```
adamstankiewicz, line 52: [curious/question] If we don't specify a `commit-message` arg here, will the PR be generated with the default commit message of "[create-pull-request] automated change" ([docs](https://github.com/marketplace/actions/create-pull-request))? When this PR merges to `master`, we'd ideally be using the commit message specified in the Lerna.json config (i.e., `chore(release): ...`).

If we're generating a PR for the changes with its own distinct commit message now instead of having Lerna generate the commit message, do we possibly need to bring the `chore(release): ...`) commit message into the `commit-message` option for `peter-evans/create-pull-request@v4` instead?

lerna.json
```diff
@@ -5,8 +5,8 @@
   "version": "independent",
   "command": {
     "publish": {
-      "allowBranch": "master",
-      "message": "chore(release): publish :tada:",
+      "allowBranch": "automation/lerna",
```
adamstankiewicz, line 8: The `allowBranch` field dictates which branches `lerna version` and/or `lerna publish` can run on. As Lerna is run after a PR is merged into `master`, `lerna version` still needs to be allowed to run on `master`. It doesn't actually make sense to have this be aligned to the generated PR branch as we aren't running `lerna version` in the CI for the PR itself (only once its merged). That said, there is the open question of how/when the Git tags and Github release would be created/pushed based on the Lerna-generated PR, that might dictate needing to run `lerna version` in some form from a PR itself.

Related, the "Publish from package.json" GH workflow (the manual step after you merge a PR) is running `lerna publish from-package ...`. As such, the `allowBranch` still needs to be set to `master` at least for the "publish" command, too.

If needed, `allowBranch` does accept a list for multiple branches as well (e.g., if we do need to run `lerna version` on the generated branch).

.github/workflows/release.yml
adamstankiewicz: @alex-sheehan-edx That's just the name of the Github Action; it represents the Github organization and repo name where the Github Action is defined: https://github.com/peter-evans/create-pull-request

lerna.json
macdiesel: I didn't read the lerna docs and made an assumption. Will go digest this now.

.github/workflows/release.yml
macdiesel: Once we get something working I will test this change and see what happens.

.github/workflows/ci.yml
```diff
@@ -4,6 +4,7 @@ on: [pull_request]
 
 jobs:
   tests:
+    if: "!(github.actor == '?AutomationGuy?' and contains(github.event.commits[0].message, 'chore(release): publish'))"
```
alex-sheehan-edx, line 7: why am I giggling at "AutomationGuy"

.github/workflows/release.yml
macdiesel: I've changed this to ${{ github.event.commits[0].message }}

.github/workflows/publish-from-package.yml
```diff
@@ -5,6 +5,7 @@ on: [workflow_dispatch]
 jobs:
   release:
     name: Release
+    if: "!(github.actor == 'github-actions[bot]' and contains(github.event.commits[0].message, 'chore(release): publish'))"
```
adamstankiewicz, line 8: I'm not sure we necessarily need to make any changes to this workflow. This one needs to be triggered manually on `master` once Lerna's `chore: publish` commit lands on `master`. There's no commit associated with the run of this workflow.

.github/workflows/ci.yml
```diff
@@ -4,6 +4,11 @@ on: [pull_request]
 
 jobs:
   tests:
+    env:
+      RUN_EVAL: ${{ !(github.actor == 'github-actions[bot]') }}
+      run: echo "$RUN_EVAL"
+    if: "!(github.actor == 'github-actions[bot]') "
```
adamstankiewicz, line 10: [question] Playing devil's advocate on this one a bit, should we run ci.yml on all commits? For example, say there's a commit from a bot one day using this `github-actions[bot]` actor bumping the version of some dependencies in various packages within the repo. We may want to ensure tests, etc. are run for those packages in this case?

macdiesel: The plan here is to match on the user and the base commit message if I can ever figure out how to get it. You're looking at a partial broken implementation here.

PR #309, opened 2023-03-21
build: upgrade frontend-platform to v3

Syed-Ali-Abbas-Zaidi:
## Description

This PR upgrades `@edx/frontend-platform` to `v3`.

### DIFF
--- a/packages/catalog-search/package.json
+++ b/packages/catalog-search/package.json
@@ -45,7 +45,7 @@
   "devDependencies": {
     "@edx/browserslist-config": "1.1.0",
     "@edx/frontend-build": "12.3.0",
-    "@edx/frontend-platform": "1.15.6",
+    "@edx/frontend-platform": "^3.6.0",
     "@edx/paragon": "19.25.1",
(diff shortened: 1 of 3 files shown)


packages/catalog-search/package.json
```diff
@@ -60,7 +60,7 @@
     "react-test-renderer": "16.14.0"
   },
   "peerDependencies": {
-    "@edx/frontend-platform": "^1.9.6 || ^2.0.0",
+    "@edx/frontend-platform": "^3.6.0",
```
adamstankiewicz, line 63: Note: this is a breaking change to remove `v1` and `v2` support for `@edx/frontend-platform` from peer dependencies. Is there anything in `catalog-search` that precludes it from using `v1` or `v2` of `@edx/frontend-platform`?

If not, we should keep these peer dependencies in place, but expand them to include v3 of `@edx/frontend-platform` to avoid a breaking change for consumers. Also note, it should include the earliest compatible v3 version of `@edx/frontend-platform` (not just the latest version).

Similar feedback for the other packages in this repo.

PR #371, opened 2024-01-30
feat: add prequery search suggestions

katrinan029:
## Description
We want to add a prequery suggestions to the learner portal using the bayesian average that was [added to algolia](https://2u-internal.atlassian.net/browse/ENT-8193). The idea being when the course search box is focused, we give a drop down box selection of suggested courses based on their review data.

Related PR in learner portal: [#935](https://github.com/openedx/frontend-app-learner-portal-enterprise/pull/935)

[JIRA](https://2u-internal.atlassian.net/browse/ENT-8194)

https://github.com/openedx/frontend-enterprise/assets/71999631/45803994-7294-427a-bdfc-c27f2d25e08c

**Merge checklist:**

**Post merge:**
- [ ] Verify Lerna created a release commit (e.g., ``chore(release): publish``) that incremented versions in relevant package.json and CHANGELOG files, and created [Git tags](https://github.com/openedx/frontend-enterprise/tags) for those versions.

### DIFF
--- a/packages/catalog-search/src/SearchBox.jsx
+++ b/packages/catalog-search/src/SearchBox.jsx
@@ -44,6 +44,7 @@ export const SearchBoxBase = ({
   enterpriseSlug,
   suggestionSubmitOverride,
   disableSuggestionRedirect,
+  isPreQueryEnabled,
 }) => {
   const { dispatch, trackingName } = useContext(SearchContext);
 
@@ -92,7 +93,6 @@ export const SearchBoxBase = ({
 
   // Track the focused element
   const focusedElement = useActiveElement();
-
   // Function to be called when the user stops typing, will fetch algolia hits for query after `DEBOUNCE_TIME_MS` has
   // elapsed
   const debounceFunc = async (query) => {
@@ -111,8 +111,8 @@ export const SearchBoxBase = ({
         // If there are no results of the suggested search, hide the empty suggestion component
         setShowSuggestions(false);
       }
-    // Display the prequery results when user clicks on search box but has not began typing
-    } else {
+    // If isPreQueryEnabled is true display the prequery results when user clicks on search box but has not began typing
+    } else if (query === '' && isPreQueryEnabled) {
       const { hits } = await index.search(query, {
         filters,
         attributesToHighlight: ['title'],
@@ -133,7 +133,7 @@ export const SearchBoxBase = ({
     if (index !== undefined && focusedElement.classList.contains(SEARCH_BOX_CLASS_NAME)) {
       debounceHandler(searchQuery);
     }
-  // Retry this method if the focused element or the search query changes
+    // Retry this method if the focused element or the search query changes
   }, [searchQuery, focusedElement]);
 
   /**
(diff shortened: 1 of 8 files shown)


packages/catalog-search/src/SearchBox.jsx
```diff
@@ -155,6 +164,9 @@ export const SearchBoxBase = ({
         value={defaultRefinement}
         onSubmit={handleSubmit}
         onClear={handleClear}
+        onFocus={(query) => {
```
alex-sheehan-edx, line 167: can I get a tldr on why we're setting the query on focus on top of on input change?

packages/catalog-search/src/SearchSuggestionItem.jsx
```diff
@@ -14,28 +15,44 @@ const SearchSuggestionItem = ({
...
-        {
-          authoringOrganization && (
-            <div className="badge badge-light ml-3 font-weight-light authoring-org-badge">
-              {authoringOrganization}
+    <div>
+      {isPreQuery ? (
+        <Link to={url} key={hit.title} className="prequery-item pr-4 d-flex flex-column" onClick={handleLinkDisable}>
+          <div className="d-flex align-items-center justify-content-start">
+            <Image className="prequery-image mr-2" src={hit.card_image_url} />
+            <div className="d-flex flex-column">
```
alex-sheehan-edx, line 24: are these two rendered components worth pulling out into their own separate components?

alex-sheehan-edx (approved):
A couple of thoughts, some suggestions/questions but otherwise looks good to me!

packages/catalog-search/src/styles/_SearchSuggestions.scss
```diff
@@ -13,53 +13,84 @@
...
+  .program-type {
     font-size: .9rem;
     margin-bottom: 0px;
   }
 
 }
 
+.prequery-item {
+  margin: 5px 10px 5px 10px;
+  color: #000;
```
kiram15, line 59: I would replace the hardcoded hex values with `$white` or `white`

packages/catalog-search/src/SearchSuggestionItem.jsx
```diff
@@ -14,28 +15,44 @@ const SearchSuggestionItem = ({
...
+              {/* eslint-disable-next-line react/no-danger, no-underscore-dangle */}
+              <div dangerouslySetInnerHTML={{ __html: hit._highlightResult.title.value }} />
+              <div className="x-small d-flex flex-column">
+                {hit.partners[0]?.name} | {hit.learning_type}
```
kiram15, line 28: Non-blocking nit: the learning type is lowercase like 'course' but in the Figma mocks it's uppercase

kiram15 (commented):
Everything looks good except a couple nits! One question is with the "view all results" at the bottom, it doesn't seem to do much cause it just takes you back to the same query (instead of going to just courses for example) is this expected?

packages/catalog-search/src/SearchBox.jsx
katrinan029: the `onChange` handles input changes in the search field, where as the `onFocus` is used to handle a user's focus on the component I.e. when the user clicks or tabs into the search field.

packages/catalog-search/src/styles/_SearchSuggestions.scss
katrinan029: this value is actually closer to black. I will leave this as is since it is the convention used in the current file.

packages/catalog-search/src/SearchSuggestionItem.jsx
katrinan029: good catch! this has been updated

packages/catalog-search/src/SearchSuggestionItem.jsx
katrinan029: I think that the components are similar/small enough to keep as one. lmk if you feel strongly otherwise and I can follow-up with another change.

PR #398, opened 2024-06-13
build: Update codecov and use token

arbrandes:
Update codecov to the latest version and start using the org-wide token for uploads.

See https://github.com/openedx/wg-frontend/issues/179


arbrandes:
Closing as per the comment here:

https://github.com/openedx/frontend-app-learner-portal-enterprise/pull/1099#discussion_r1657069712

arbrandes:
Actually, this is not just flipping fail_ci_if_true, but actually upgrading Codecov.  Reopening.  @adamstankiewicz?

PR #435, opened 2025-02-24
feat: support React 18 in `@edx/frontend-enterprise-hotjar`

adamstankiewicz:
**Merge checklist:**

**Post merge:**
- [ ] Follow the [release steps in the README documentation](../README.rst#versioning-and-releases). Verify Lerna's release commit (e.g., ``chore(release): publish new versions``) that incremented versions in relevant package.json and CHANGELOG files, and created [Git tags](https://github.com/openedx/frontend-enterprise/tags) for those versions is on ``master`` (**Important: ensure the Git tags are for the correct commit SHA**).

### DIFF
--- a/packages/hotjar/package.json
+++ b/packages/hotjar/package.json
@@ -19,7 +19,7 @@
     "clean": "make clean",
     "build": "make build",
     "i18n_extract": "BABEL_ENV=i18n fedx-scripts babel src --quiet > /dev/null",
-    "lint": "fedx-scripts eslint --ext .js --ext .jsx --ext .ts -ext .tsx .",
+    "lint": "fedx-scripts eslint --ext .js --ext .jsx --ext .ts --ext .tsx .",
     "lint:fix": "npm run lint -- --fix",
     "test": "fedx-scripts jest --coverage --passWithNoTests",
     "snapshot": "npm run test -- --updateSnapshot",
--- a/packages/hotjar/package.json
+++ b/packages/hotjar/package.json
@@ -11,7 +11,7 @@
       "patterns": [
         "src"
       ],
-      "extensions": "js,jsx"
+      "extensions": "js,jsx,ts,tsx"
     }
   },
   "scripts": {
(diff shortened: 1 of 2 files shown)


packages/hotjar/package.json
```diff
@@ -37,15 +37,15 @@
...
-    "@edx/browserslist-config": "1.5.0",
-    "@openedx/frontend-build": "14.3.1",
-    "react": "17.0.2",
-    "react-dom": "17.0.2",
-    "react-router-dom": "6.29.0"
+    "@edx/browserslist-config": "^1.5.0",
+    "@openedx/frontend-build": "^14.3.1",
+    "react": "^18.3.1",
+    "react-dom": "^18.3.1",
+    "react-router-dom": "^6.29.0"
```
brobro10000, line 44: Should these packages be pinned?

brobro10000 (approved):
Approved to unblock but curious if the versions of the updated packages should be pinned?

packages/hotjar/package.json
adamstankiewicz: based on some recent conversations [here](https://github.com/openedx/frontend-build/pull/625#discussion_r1963996106), it's likely the convention of pinning NPM dependencies will change where we'll want to typically rely on `^`. Opting to proactively change it for this repo while we're here.

brobro10000: This is good context on the newly minted convention that should work will with keeping our MFE packages up to date as well. 

I share the same sentiment that package updates can be prefixed WITH  `^..`, but would probably be worth sharing the information with other engineers since the pinned package version has been engrained in many peoples workflow.

adamstankiewicz: Definitely. Largely holding off on a broader inform until it gets into a more org-wide ADR more formally; not sure if that's something Brian was planning to take on yet; if not, we could drive that forward, too.

Regardless, it'll actually make dev easier when installing/updating deps since NPM by default uses `^`; usually takes effort to remember to undo the `^` and re-run `npm i` to pin it when doing something like `npm i @openedx/paragon@22`.

PR #462, opened 2025-04-23
fix: include only content type program and course in search suggestions

katrinan029:
https://2u-internal.atlassian.net/browse/ENT-10331

Update Algolia filter to only include programs and course content types in search suggestions because we are currently not supporting videos and pathways.

**Merge checklist:**

**Post merge:**
- [ ] Follow the [release steps in the README documentation](../README.rst#versioning-and-releases). Verify Lerna's release commit (e.g., ``chore(release): publish new versions``) that incremented versions in relevant package.json and CHANGELOG files, and created [Git tags](https://github.com/openedx/frontend-enterprise/tags) for those versions is on ``master`` (**Important: ensure the Git tags are for the correct commit SHA**).

### DIFF
--- a/packages/catalog-search/src/SearchHeader.jsx
+++ b/packages/catalog-search/src/SearchHeader.jsx
@@ -34,6 +34,12 @@ const SearchHeader = ({
     searchQueryFromRefinements = refinements.q;
   }
 
+  // Add filter to only include course and program content_types
+  // as we are currently not supporting videos and pathways
+  const modifiedFilters = filters
+    ? `${filters} AND (content_type:"course" OR content_type:"program")`
+    : '(content_type:"course" OR content_type:"program")';
+
   return (
     <div
       data-testid={searchHeaderTestId}
@@ -55,7 +61,7 @@ const SearchHeader = ({
                 headerTitle={headerTitle}
                 hideTitle={hideTitle}
                 index={index}
-                filters={filters}
+                filters={modifiedFilters}
                 enterpriseSlug={slug}
                 suggestionSubmitOverride={suggestionSubmitOverride}
                 disableSuggestionRedirect={disableSuggestionRedirect}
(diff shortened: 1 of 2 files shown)


packages/catalog-search/src/SearchHeader.jsx
```diff
@@ -34,6 +34,9 @@ const SearchHeader = ({
     searchQueryFromRefinements = refinements.q;
   }
 
+  // Add filter to exclude video content type
+  const modifiedFilters = filters ? `${filters} AND NOT content_type:"video"` : "NOT content_type:\"video\"";
```
adamstankiewicz, line 38: I might flip this around and consider only _including_ the content types officially supported by `SearchSuggestions` (courses & programs), rather than excluding unsupported content types.

Rationale being, pathways could also potentially be surfaced, but the `SearchSuggestions` component doesn't support pathways currently. 

By having these filters only include courses & pathways for now prevents similar regressions from occurring if/when new content types are introduced in the future (e.g., where we wouldn't need to make any changes to _exclude_ the new content type).

packages/catalog-search/src/SearchSuggestions.jsx
```diff
@@ -35,7 +35,10 @@ const SearchSuggestions = ({
 
   if (preQueryHits) {
     preQueryHits.forEach((hit) => {
-      preQuerySuggestions.push(hit);
+      // or filter out here?
+      if (hit.content_type !== 'video') {
+        preQuerySuggestions.push(hit);
+      }
```
adamstankiewicz, line 41: I think you're on the right path with modifying the filters passed to the `SearchBox` component, so that it is passed along to both [here](https://github.com/openedx/frontend-enterprise/blob/21ce6f6fc1a6e9abd3510d64226c4e8c2921aea7/packages/catalog-search/src/SearchBox.jsx#L111) and [here](https://github.com/openedx/frontend-enterprise/blob/21ce6f6fc1a6e9abd3510d64226c4e8c2921aea7/packages/catalog-search/src/SearchBox.jsx#L126).

tl;dr; add the addtl filters as far upstream as possible so that `hits` never include unsupported content types to begin with :)

adamstankiewicz (approved):

